PR #69, opened 2022-09-27
Add multilabel support to SetFitModel

mpangrazzi:
I'm trying to integrate the multilabel example script into the `SetFitModel` class.

Basically you will need to initialize the model in this way, passing the `multi_target_strategy` option (which will create the specific head classifier):

```python

model = SetFitModel.from_pretrained("paraphrase-mpnet-base-v2", multi_target_strategy="one-vs-rest")
```

At the moment the `train()` method seems to work, but or I'm probably not using well the `sentence_pairs_generation_multilabel` or I don't know why the `predict` / `predict_proba` results are not as good as the singelabel model. 

I haven't found many tests on the existing code, so if you have any other hint it will be very appreciated!


src/setfit/modeling.py
```diff
@@ -29,10 +31,11 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
 class SetFitModel(PyTorchModelHubMixin):
     """A SetFit model with integration to the Hugging Face Hub."""
 
-    def __init__(self, model_body=None, model_head=None):
+    def __init__(self, model_body=None, model_head=None, multi_target_strategy="one-vs-rest"):
```
lewtun, line 34: I think we should set this to `None` so we can keep logistic regression as the default:
```suggestion
    def __init__(self, model_body=None, model_head=None, multi_target_strategy=None):
```

src/setfit/modeling.py
```diff
@@ -95,8 +99,21 @@ def _from_pretrained(
         if model_head_file is not None:
             model_head = joblib.load(model_head_file)
         else:
-            model_head = LogisticRegression()
-        return SetFitModel(model_body=model_body, model_head=model_head)
+            if multi_target_strategy:
```
lewtun, line 102: nit: we generally prefer to be explicit :)
```suggestion
            if multi_target_strategy is not None:
```

tests/test_modeling.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+
+from setfit.modeling import sentence_pairs_generation, sentence_pairs_generation_multilabel
+
+
+def test_sentence_pairs_generation():
```
lewtun, line 6: Amazing - thanks for adding tests!!

tests/test_setfit_model.py
```diff
@@ -0,0 +1,42 @@
+from sentence_transformers import SentenceTransformer
```
lewtun, line 1: I suggest moving these tests into `test_modeling.py` - it makes it a bit easier to find tests that are grouped by filename this way

lewtun (commented):
Thank you so much for adding this clean PR @mpangrazzi !

Regarding performance, @fgbelidji told me that he tested his script on the `go_emotions` dataset: https://huggingface.co/datasets/go_emotions

Would you like to run your notebook on that and see how it fares?

I've also left a few nits, but otherwise this is looking great. Re tests, the ones you've added look great ❤️

lewtun:
FYI I ran the `run_fewshot_multilabel.py` script in `scripts/setfit` and here's some reference numbers on `go_emotions`:

| Num samples / class | Accuracy |
| :---: | :---: |
| 8 | 18.2 +/- 2.2 |
| 64 | 31.0 +/- 0.7 |

Looking at some Hub models that have been trained on this dataset, these numbers look about right compared to training on the full corpus (typically gives accuracy of ~55%)

mpangrazzi:
Wise suggestions, thank you! I will apply them. 

Also thanks for the accuracy results, after the refactoring I will try to run the same using `SetFitModel` and compare them

tests/test_modeling.py
mpangrazzi: 🙏

lewtun:
Hey @mpangrazzi we've just added some small changes to the `SetFitTrainer` class, so you'll need to resolve some merge conflicts before we can do a final review :)

One thing that occurred to me: the `metric` arg in `SetFit` trainer assumes that metrics can be loaded / computed with a single keyword like `accuracy` etc. For the multi-label case, we might need to rethink a bit this restriction on passing a `str` and allow users to pass e.g. `sklearn` metrics directly

src/setfit/trainer.py
```diff
@@ -101,9 +110,16 @@ def train(self):
     def evaluate(self):
         """Computes the metrics for a given classifier."""
         metric_fn = evaluate.load(self.metric)
+
         x_test = self.eval_dataset["text"]
-        y_test = self.eval_dataset["label"]
+
+        if self.model.multi_target_strategy:
+            y_test = self.train_dataset.remove_columns("text").to_pandas().values
```
sanderland, line 117: eval_dataset

notebooks/text-classification_multilabel.ipynb
```diff
@@ -0,0 +1,336 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "126e1d86",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Select N examples per class (8 in this case)\n",
+    "train_ds = dataset[\"train\"].shuffle(seed=42).select(range(8 * 2))\n",
```
sanderland, line 143: this assumes 2 classes, while the dataset has more, could do
```
num_labels = dataset["train"].shape[1] - 1
train_ds = dataset["train"].shuffle(seed=42).select(range(8 * num_labels))
```

sanderland:
> One thing that occurred to me: the `metric` arg in `SetFit` trainer assumes that metrics can be loaded / computed with a single keyword like `accuracy` etc. For the multi-label case, we might need to rethink a bit this restriction on passing a `str` and allow users to pass e.g. `sklearn` metrics directly

```
metric_fn = evaluate.load(self.metric, config_name="multilabel" if self.model.multi_target_strategy else None)
```
Makes accuracy work at least.

lewtun:
> > One thing that occurred to me: the `metric` arg in `SetFit` trainer assumes that metrics can be loaded / computed with a single keyword like `accuracy` etc. For the multi-label case, we might need to rethink a bit this restriction on passing a `str` and allow users to pass e.g. `sklearn` metrics directly
> 
> ```
> metric_fn = evaluate.load(self.metric, config_name="multilabel" if self.model.multi_target_strategy else None)
> ```
> 
> Makes accuracy work at least.

Indeed, this also works for other classification metrics like `f1`, `recall`, and `precision`. I suggest we adopt this proposal until we have a really good reason to add more complex metrics :)

mpangrazzi:
Thank you. I've resolved conflicts, applied suggestions and added a `coverage` script. Still have to test results yet, but I guess it could be enough for now

notebooks/text-classification_multilabel.ipynb
```diff
@@ -0,0 +1,336 @@
...
+    "\n",
+    "from setfit import SetFitModel, SetFitTrainer"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cbd3e642",
+   "metadata": {},
+   "source": [
+    "We start by loading a dataset to work with. In this example we'll load the \"emotion\" dataset already on the HF hub."
```
lewtun, line 72: The `emotion` dataset isn't a true multilabel dataset, so what do you think about trying something like `lex_glue`?

src/setfit/trainer.py
```diff
@@ -119,7 +119,11 @@ def train(self):
             logger.info("Applying column mapping to training dataset")
             self.train_dataset = self._apply_column_mapping(self.train_dataset, self.column_mapping)
         x_train = self.train_dataset["text"]
-        y_train = self.train_dataset["label"]
+
+        if self.model.multi_target_strategy:
+            y_train = self.train_dataset.remove_columns("text").to_pandas().values
```
lewtun, line 124: I think this won't work for generic datasets where one has more than two columns. I know this logic comes from the multilabel script, but I wonder if we can make use of the column mapping added in #75 and always derive `x_train` and `y_train` from `text` and `label` columns?

In other words, can we drop this if/else statement alltogether?

src/setfit/trainer.py
```diff
@@ -186,15 +195,24 @@ def train(self):
...
             self.eval_dataset = self._apply_column_mapping(self.eval_dataset, self.column_mapping)
-        metric_fn = evaluate.load(self.metric)
+
+        metric_config = "multilabel" if self.model.multi_target_strategy is not None else None
+        metric_fn = evaluate.load(self.metric, config_name=metric_config)
+
         x_test = self.eval_dataset["text"]
-        y_test = self.eval_dataset["label"]
+
+        if self.model.multi_target_strategy:
```
lewtun, line 208: Similar comment as above - the column mapping logic in #75 should ensure that the `eval_dataset` always has a `text` and `label` column, which means we can remove this if/else logic

tests/test_modeling.py
```diff
@@ -0,0 +1,74 @@
...
+    for _ in range(n_iterations):
+        pairs = sentence_pairs_generation_multilabel(sentences, labels, pairs)
+
+    assert len(pairs) == 12
+    assert pairs[0].texts == ["sent 1", "sent 1"]
+    assert pairs[0].label == 1.0
+
+
+def test_setfit_model_body():
+    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")
```
lewtun, line 42: let's use a tiny model to save some bandwidth :)
```suggestion
    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
```

tests/test_modeling.py
```diff
@@ -0,0 +1,74 @@
...
+
+    assert type(model.model_body) is SentenceTransformer
+
+
+def test_setfit_default_model_head():
+    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")
```
lewtun, line 48:
```suggestion
    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
```

tests/test_modeling.py
```diff
@@ -0,0 +1,74 @@
...
+
+    assert type(model.model_head) is LogisticRegression
+
+
+def test_setfit_multilabel_one_vs_rest_model_head():
+    model = SetFitModel.from_pretrained(
+        "sentence-transformers/paraphrase-mpnet-base-v2", multi_target_strategy="one-vs-rest"
```
lewtun, line 55:
```suggestion
        "sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="one-vs-rest"
```

tests/test_modeling.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    assert type(model.model_head) is OneVsRestClassifier
+
+
+def test_setfit_multilabel_multi_output_classifier_model_head():
+    model = SetFitModel.from_pretrained(
+        "sentence-transformers/paraphrase-mpnet-base-v2", multi_target_strategy="multi-output"
```
lewtun, line 63:
```suggestion
        "sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="multi-output"
```

tests/test_modeling.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    assert type(model.model_head) is MultiOutputClassifier
+
+
+def test_setfit_multilabel_classifier_chain_classifier_model_head():
+    model = SetFitModel.from_pretrained(
+        "sentence-transformers/paraphrase-mpnet-base-v2", multi_target_strategy="classifier-chain"
```
lewtun, line 71:
```suggestion
        "sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="classifier-chain"
```

notebooks/text-classification_multilabel.ipynb
```diff
@@ -0,0 +1,336 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d56e5027",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "!echo CUDA_VISIBLE_DEVICES"
```
lewtun, line 231: I think we can delete this - must be an artifact from the original notebook :)

lewtun (commented):
Thanks for iterating on this great addition @mpangrazzi! I've left a few small comments, but think this is getting close to being merged 🔥

mpangrazzi:
@lewtun I forgot about that notebook :) Will fix it soon. Ok also for applying column mapping logic

lewtun:
> @lewtun I forgot about that notebook :) Will fix it soon. Ok also for applying column mapping logic

FYI I recently revamped the original notebook to make it a bit more structured - feel free to copy-paste + tweak for the multilabel case!

lewtun:
Hey @mpangrazzi if creating the notebook is slowing you down too much, feel free to address the comments on the source code and we can add the notebook later!

src/setfit/trainer.py
mpangrazzi: I have a question here: 

I like the idea of expecting always `text` and `label` columns also for multilabel case, but when using datasets like `lex_glue` (`eurlex`) labels are non-binary sequences (e.g. `[ 28, 32, 33, 91, 96, 97 ]`) while `sentence_pairs_generation_multilabel` expects binary sequences. How do you suggest to handle them?

sanderland: The huggingface libraries typically leave binarization to the user (using `map` and the like), and handling binarization in a generic way can get [rather involved](https://github.com/chatdesk/grouphug/blob/master/grouphug/dataset_formatter.py#L274).
Could a reasonable intermediate be to allow a list for the `label` value in columns mapping? That way you can handle `go_emotions` datasets at least somewhat cleanly.

lewtun (commented):
Thanks for iterating on the PR @mpangrazzi - it's looking great! 

There's a minor merge conflict to be resolved and I think we can delete the notebook for now and leave that as a separate addition. 

After that, we should be good to merge this nice feature!

mpangrazzi:
@lewtun should be everything ok now. Agree to delete notebook and put in on a separate request!

lewtun (approved):
Wonderful, thanks for the final iteration - this looks great! I'll add a note in the README so people know about this feature :)

PR #82, opened 2022-10-05
Fix column mapping checks

lewtun:
As reported by @osanseviero, the recent label mapping change introduced a bug in the blog post, where one cannot train a SetFit model without specifying the mapping explicitly.

I implemented a unit test to reproduce the error and then fixed the source.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ on:
 jobs:
 
   test_sampling:
-    name: Test sampling functions
+    name: Run unit tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code


src/setfit/trainer.py
```diff
@@ -78,7 +78,7 @@ def _validate_column_mapping(self, dataset: "Dataset") -> None:
         """
         required_columns = set(["text", "label"])
         column_names = set(dataset.column_names)
-        if self.column_mapping is None and not column_names.issubset(required_columns):
+        if self.column_mapping is None and not required_columns.issubset(column_names):
```
lewtun, line 81: This was the offending line that triggered the error

tests/test_trainer.py
```diff
@@ -1,44 +1,53 @@
...
-    with pytest.raises(ValueError):
-        trainer._validate_column_mapping(trainer.train_dataset)
-
-
-def test_column_mapping_with_missing_text(setup_trainer):
-    trainer = setup_trainer
-    trainer.column_mapping = {"label_new": "label"}
-    with pytest.raises(ValueError):
-        trainer._validate_column_mapping(trainer.train_dataset)
+class SetFitTrainerTest(TestCase):
```
lewtun, line 10: I took the liberty of refactoring the tests with unittest, which allows for better grouping

lewtun:
Alright, merging in the interest of making a patch release so the blog actually works :)

PR #87, opened 2022-10-06
Add model_init support to SetFitTrainer

Mouhanedg56:
This PR is to add `model_init` arg to `SetFitTrainer`.
If provided, each call to `SetFitTrainer.train` will start from a new instance of the model as given by this function:

```python
from setfit.modeling import SetFitModel
from setfit.trainer import SetFitTrainer

def get_model():
    model_name = "sentence-transformers/paraphrase-albert-small-v2"
    return SetFitModel.from_pretrained(model_name)

dataset = Dataset.from_dict(
            {"text_new": ["a", "b", "c"], "label_new": [0, 1, 2], "extra_column": ["d", "e", "f"]}
        )
trainer = SetFitTrainer(
            model_init=get_model,
            train_dataset=dataset,
            eval_dataset=dataset,
            num_iterations=1,
            column_mapping={"text_new": "text", "label_new": "label"},
        )
```

From now on, `SetFitTrainer` requires either a `model` or `model_init` argument, but not both.
I added a unit test on `test_trainer.py` to test initiating a trainer using `model_init`.

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -81,10 +81,10 @@ class SetFitTrainer:
                 self.model_init = model_init
                 model = self.call_model_init()
             else:
-                raise RuntimeError("`Trainer` requires either a `model` or `model_init` argument")
+                raise RuntimeError("`SetFitTrainer` requires either a `model` or `model_init` argument")
         else:
             if model_init is not None:
-                raise RuntimeError("`Trainer` requires either a `model` or `model_init` argument, but not both")
+                raise RuntimeError("`SetFitTrainer` requires either a `model` or `model_init` argument, but not both")
 
             self.model_init = model_init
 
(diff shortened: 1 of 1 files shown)


src/setfit/trainer.py
```diff
@@ -113,7 +131,27 @@ def _apply_column_mapping(self, dataset: "Dataset", column_mapping: Dict[str, st
         )
         return dataset
 
+    def call_model_init(self):
```
lewtun, line 134: Looking ahead to including trials from e.g. `optuna`, should we extend the function signature to accept a `trial` arg?

tests/test_trainer.py
```diff
@@ -12,6 +12,25 @@ def setUp(self):
         self.model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
         self.num_iterations = 1
 
+    def test_trainer_works_with_model_init(self):
```
lewtun, line 15: Great to see this tested!

lewtun (commented):
Thanks for adding this feature @Mouhanedg56 🔥 !

I have one small question about the function signature of the init, but otherwise this is looking great :)

src/setfit/trainer.py
Mouhanedg56: Yes we should but I think it'd be better to do that on hyperparameter search PR once the trail is implemented. What do you think?

lewtun: Sure, that sounds good to me! Let's merge this then :)

lewtun (approved):

PR #100, opened 2022-10-12
Hyperparameter optimization fixes + notebook

sanderland:
* fixes default value of 'minimize' which is not suitable for the default objective of accuracy
* adds a notebook for hyperparameter search
* adds its own copy of BestRun which doesn't suggest models are saved (since I think they are not) and saves the study object for inspection.
  *  The latter could be added upstream in transformers as well if there is interest.

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -145,7 +145,7 @@ class SetFitTrainer:
 
         Args:
             params (`Dict[str, Any]`): The parameters, usually from `BestRun.hyperparameters`
-            final_model (`bool`): If True, replace the model_init function with a fixed model based on the parameters.
+            final_model (`bool`, *optional*, defaults to `False`): If `True`, replace the `model_init()` function with a fixed model based on the parameters.
         """
         for key, value in params.items():
             if hasattr(self, key):


src/setfit/trainer.py
```diff
@@ -303,7 +297,7 @@ def hyperparameter_search(
         hp_space: Optional[Callable[["optuna.Trial"], Dict[str, float]]] = None,
         compute_objective: Optional[Callable[[Dict[str, float]], float]] = None,
         n_trials: int = 10,
-        direction: str = "minimize",
+        direction: str = "maximize",
```
Mouhanedg56, line 300: Thanks for updating this. Setfit's default metrics is accuracy, not the same as transformers default metric which is loss. So direction default value should be `maximize`.

src/setfit/utils.py
```diff
@@ -131,3 +131,24 @@ def summary(self) -> None:
...
+        hyperparameters (`Dict[str, Any]`):
+            The hyperparameters picked to get this run.
+        backend (`Any`):
+            The relevant internal object used for optimization. For optuna this is the `study` object.
+    """
+
+    run_id: str
+    objective: float
+    hyperparameters: Dict[str, Any]
+    backend: Any
```
Mouhanedg56, line 154:
```suggestion
 backend: Any = None
```

src/setfit/trainer.py
```diff
@@ -109,7 +103,7 @@ def _validate_column_mapping(self, dataset: "Dataset") -> None:
         """
         Validates the provided column mapping against the dataset.
         """
-        required_columns = set(["text", "label"])
+        required_columns = {"text", "label"}
```
lewtun, line 106: Nice catch :)

src/setfit/utils.py
```diff
@@ -131,3 +131,24 @@ def summary(self) -> None:
         Prints summary of all benchmarks performed.
         """
         self.print(f"\n{'#' * 30}\nBenchmark Summary:\n{'#' * 30}\n\n{self.summary_msg}")
+
+
+class BestRun(NamedTuple):
+    """
+    The best run found by an hyperparameter search (see [`~Trainer.hyperparameter_search`]).
```
lewtun, line 138: should this refer to the `SetFitTrainer` or the `Trainer` from `transformers`? If the former, I'd use this:
```suggestion
    The best run found by an hyperparameter search (see [`~SetFitTrainer.hyperparameter_search`]).
```

lewtun (approved):
Thanks for fixing the default search direction and adding this really nice notebook @sanderland 😍 !

The PR looks good and I left a few small nits

notebooks/text-classification_optuna.ipynb
```diff
@@ -0,0 +1,401 @@
+{
```
lewtun, line 1: <p>Very cool plot!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/huggingface/setfit/pull/100/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

lewtun, line 1: <p>Since the original trainer uses <code>model_init</code> I think we can simplify this step to:</p><p> </p><pre class="ql-syntax" spellcheck="false">for k,v in best_run.hyperparameters.items():
    setattr(trainer, k, v) 
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/huggingface/setfit/pull/100/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

sanderland: <p>That would not catch multi_target_strategy. Will have a look if I can make it a bit cleaner</p>

sanderland: @Mouhanedg56 @lewtun Refactored a bit to not have code duplication there. I think the interface is quite nice, however:

* It is diverging a little from transformers. In a good way I think, but could be surprising.
* Tests need a little refactoring if we go with this.

src/setfit/trainer.py
```diff
@@ -146,40 +140,53 @@ def _apply_column_mapping(self, dataset: "Dataset", column_mapping: Dict[str, st
         )
         return dataset
 
+    def apply_hyperparameters(self, params: Dict[str, Any], final_model: bool = False):
+        """Applies a dictionary of hyperparameters to both the trainer and the model
+
+        Args:
+            params (`Dict[str, Any]`): The parameters, usually from `BestRun.hyperparameters`
+            final_model (`bool`): If True, replace the model_init function with a fixed model based on the parameters.
```
lewtun, line 148: turbo nit:

```suggestion
            final_model (`bool`, *optional*, defaults to `False`): If `True`, replace the `model_init()` function with a fixed model based on the parameters.
```

notebooks/text-classification_optuna.ipynb
lewtun: Ah good point re `multi_target_strategy`!

I'm happy to diverge from the `transformers.Trainer` since it's only used as inspiration :) The current API in your PR looks neat to me, so let's use that + a small refactoring of the tests

lewtun (commented):
Thanks for iterating on the API @sanderland 💪 !

I left a nit on the docstring - maybe we can also update the example in the `README` to show the `apply_hyperparameters()` method?

notebooks/text-classification_optuna.ipynb
Mouhanedg56: Great refactoring. The usage now is much easier than the previous version.

sanderland:
Should be all done @lewtun

lewtun (commented):
Thanks for iterating, this is great!

PR #112, opened 2022-10-19
Support for the differentiable head

blakechi:
This pull request add supports as mentioned in the issue #8.
- added `SetFitHead`, which inherited from Sentence Transformers' [Dense](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/models/Dense.py) to make APIs consistent
- integrated `SetFitHead` to `SetFitModel`
- integrated `SetFitHead` to `SetFitTrainer` (sklearn-based head still works and usage remains the same)
- added new tests to test `SetFitHead` (tested initialization for single/multiple targets, forward, backward)
- added new APIs for `SetFitTrainer`: `trainer.freeze()` and `trainer.unfreeze(keep_body_frozen)`

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -56,6 +56,11 @@ class SetFitTrainer:
             [`~SetTrainer.model_init`] function to instantiate the model if it has some randomly initialized parameters.
         column_mapping (`Dict[str, str]`, *optional*):
             A mapping from the column names in the dataset to the column names expected by the model. The expected format is a dictionary with the following format: {"text_column_name": "text", "label_column_name: "label"}.
+        use_amp (`bool`, *optional*, defaults to `False`):
+            Use Automatic Mixed Precision (AMP). Only for Pytorch >= 1.6.0
+        warmup_proportion (`float`, *optional*, defaults to `0.1`):
+            Proportion of the warmup in the total training steps.
+            Must be greater than or equal to 0.0 and less than or equal to 1.0.
     """
 
     def __init__(
@@ -72,7 +77,13 @@ class SetFitTrainer:
         batch_size: int = 16,
         seed: int = 42,
         column_mapping: Dict[str, str] = None,
+        use_amp: bool = False,
+        warmup_proportion: float = 0.1,
     ):
+        if (warmup_proportion < 0.0) or (warmup_proportion > 1.0):
+            raise ValueError(
+                f"warmup_proportion must be greater than or equal to 0.0 and less than or equal to 1.0! But it was: {warmup_proportion}"
+            )
 
         self.train_dataset = train_dataset
         self.eval_dataset = eval_dataset
@@ -84,6 +95,8 @@ class SetFitTrainer:
         self.batch_size = batch_size
         self.seed = seed
         self.column_mapping = column_mapping
+        self.use_amp = use_amp
+        self.warmup_proportion = warmup_proportion
 
         if model is None:
             if model_init is not None:
@@ -304,6 +317,7 @@ class SetFitTrainer:
                 elif self.loss_class is SupConLoss:
                     train_loss = self.loss_class(model=self.model)
                 else:
+
                     train_loss = self.loss_class(
                         model=self.model,
                         distance_metric=BatchHardTripletLossDistanceFunction.cosine_distance,
@@ -326,7 +340,7 @@ class SetFitTrainer:
 
                 train_dataloader = DataLoader(train_examples, shuffle=True, batch_size=batch_size)
                 train_loss = self.loss_class(self.model.model_body)
-                train_steps = len(train_dataloader)
+                train_steps = len(train_dataloader) * num_epochs
 
             logger.info("***** Running training *****")
             logger.info(f"  Num examples = {len(train_examples)}")
@@ -334,7 +348,7 @@ class SetFitTrainer:
             logger.info(f"  Total optimization steps = {train_steps}")
             logger.info(f"  Total train batch size = {batch_size}")
 
-            warmup_steps = math.ceil(train_steps * 0.1)
+            warmup_steps = math.ceil(train_steps * self.warmup_proportion)
             self.model.model_body.fit(
                 train_objectives=[(train_dataloader, train_loss)],
                 epochs=num_epochs,
@@ -342,6 +356,7 @@ class SetFitTrainer:
                 optimizer_params={"lr": learning_rate},
                 warmup_steps=warmup_steps,
                 show_progress_bar=True,
+                use_amp=self.use_amp,
             )
 
         if not is_differentiable_head or not self._freeze:
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<img src="assets/setfit.png">
+<img src="https://raw.githubusercontent.com/huggingface/setfit/main/assets/setfit.png">
 
 <p align="center">
     🤗 <a href="https://huggingface.co/setfit" target="_blank">Models & Datasets</a> | 📖 <a href="https://huggingface.co/blog/setfit" target="_blank">Blog</a> | 📃 <a href="https://arxiv.org/abs/2209.11055" target="_blank">Paper</a>
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,10 @@
 # Lint as: python3
-"""Setup for repository"""
+from pathlib import Path
 
 from setuptools import find_packages, setup
 
 
-DOCLINES = __doc__.split("\n")
+README_TEXT = (Path(__file__).parent / "README.md").read_text(encoding="utf-8")
 
 INTEGRATIONS_REQUIRE = ["optuna"]
 
@@ -27,8 +27,9 @@ EXTRAS_REQUIRE["dev"] = combine_requirements([k for k in EXTRAS_REQUIRE])
 setup(
     name="setfit",
     version="0.4.0.dev0",
-    description=DOCLINES[0],
-    long_description="\n".join(DOCLINES[2:]),
+    description="Efficient few-shot learning with Sentence Transformers",
+    long_description=README_TEXT,
+    long_description_content_type="text/markdown",
     author="SetFit",
     author_email="lewis@huggingface.co",
     url="https://github.com/SetFit/setfit",
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -137,6 +137,67 @@ def create_fewshot_splits_multilabel(dataset: Dataset, sample_sizes: List[int]) 
     return splits_ds
 
 
+def add_templated_examples(
+    dataset: Dataset,
+    candidate_labels: List[str],
+    template: str = "This sentence is {}",
+    sample_size: int = 2,
+    text_column: str = "text",
+    label_column: str = "label",
+) -> Dataset:
+    """Adds templated examples to a Dataset.
+
+    The Dataset is assumed to have a text column with the name `text_column` and a
+    label column with the name `label_column`, which contains one-hot or multi-hot
+    encoded label sequences.
+
+    Args:
+        dataset (`Dataset`): The Dataset to add templated examples to.
+        candidate_labels (`List[str]`): This list of candidate labels to be fed into
+            the template to construct examples. This should align with the
+            `label_column_name` column of `dataset`.
+        template (`str`, *optional*, defaults to `"This sentence is {}"`): The template
+            used to turn each label into a synthetic training example. This template
+            must include a {} for the candidate label to be inserted into the template.
+            For example, the default template is "This sentence is {}." With the
+            candidate label "sports", this would produce an example
+            "This sentence is sports".
+        sample_size (`int`, *optional*, defaults to 2): The number of examples to
+            make for each candidate label.
+        text_column (`str`, *optional*, defaults to `"text"`): The name of the column
+            containing the text of the examples.
+        label_column (`str`, *optional*, defaults to `"label"`): The name of the column
+            containing the labels of the examples.
+
+    Returns:
+        `Dataset`: A copy of the input Dataset with templated examples added.
+
+    Raises:
+        `ValueError`: If the input Dataset is not empty and one or both of the
+            provided column names are missing.
+    """
+    required_columns = {text_column, label_column}
+    column_names = set(dataset.column_names)
+    if column_names:
+        missing_columns = required_columns.difference(column_names)
+        if missing_columns:
+            raise ValueError(f"The following columns are missing from the input dataset: {missing_columns}.")
+
+    empty_label_vector = [0] * len(candidate_labels)
+
+    for label_id, label_name in enumerate(candidate_labels):
+        label_vector = empty_label_vector
+        label_vector[label_id] = 1
+        example = {
+            text_column: template.format(label_name),
+            label_column: label_vector,
+        }
+        for _ in range(sample_size):
+            dataset = dataset.add_item(example)
+
+    return dataset
+
+
 class SetFitDataset(TorchDataset):
     """SetFitDataset
 
--- a/tests/test_add_templated_examples.py
+++ b/tests/test_add_templated_examples.py
@@ -1,0 +1,67 @@
+import pytest
+from datasets import Dataset
+
+from setfit.data import add_templated_examples
+
+
+@pytest.fixture
+def empty_dataset():
+    return Dataset.from_dict({})
+
+
+@pytest.fixture
+def dataset():
+    return Dataset.from_dict(
+        {
+            "text": ["label-0 text", "label-1 text"],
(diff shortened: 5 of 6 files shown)


src/setfit/modeling.py
```diff
@@ -23,6 +25,43 @@
 MODEL_HEAD_NAME = "model_head.pkl"
 
 
+class SetFitDataset(Dataset):
+    def __init__(self, x, y, tokenizer, max_length=32):
```
lewtun, line 29: Small nit: could we add some types here? For things like `tokenizer`, we can import them with `typing.TYPE_CHECKING` and then wrap the type in quotes " ... "

src/setfit/modeling.py
```diff
@@ -23,6 +25,43 @@
...
+        self.x = x
+        self.y = y
+        self.tokenizer = tokenizer
+        self.max_length = max_length
+
+    def __len__(self):
+        return len(self.x)
+
+    def __getitem__(self, idx):
```
lewtun, line 38: nit:
```suggestion
    def __getitem__(self, idx: int):
```

src/setfit/modeling.py
```diff
@@ -33,6 +72,92 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
             self.model._modules["2"] = models.Normalize()
 
 
+class SetFitHead(models.Dense):
+    """
+    A SetFit head that supports binary/multi-classes logistic regression
```
lewtun, line 77: turbo nit:
```suggestion
    A SetFit head that supports binary and multi-class classification
```

src/setfit/modeling.py
```diff
@@ -33,6 +72,92 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+    for end-to-end training.
+
+    To be compatible with Sentence Transformers, we inherit `Dense` from:
+    https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/models/Dense.py
+
+    Args:
+        in_features (`int`, *optional*):
+            The embedding dimension from the output of the SetFit body. If ignore, will use LazyLinear.
```
lewtun, line 85:
```suggestion
            The embedding dimension from the output of the SetFit body. If`None`, defaults to `LazyLinear`.
```

src/setfit/modeling.py
```diff
@@ -33,6 +72,92 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+
+    def get_config_dict(self):
+        return {
+            "in_features": self.in_features,
+            "out_features": self.out_features,
+            "temperature": self.temperature,
+            "bias": self.bias,
+        }
+
+    def __repr__(self):
```
lewtun, line 157: Nice!

src/setfit/modeling.py
```diff
@@ -41,13 +166,93 @@ def __init__(self, model_body=None, model_head=None, multi_target_strategy=None)
...
+        self,
+        x_train: List[str],
+        y_train: List[int],
+        num_epochs: Optional[int] = None,
+        batch_size: Optional[int] = None,
+        learning_rate: Optional[float] = None,
+        body_learning_rate: Optional[float] = None,
+        l2_weight: Optional[float] = None,
+    ):
+        if isinstance(self.model_head, nn.Module):  # train with pyTorch
```
lewtun, line 184: Very elegant way to preserve backwards compatibility!

src/setfit/modeling.py
```diff
@@ -33,6 +72,92 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+        outputs = None
+        if self.out_features == 1:  # only has one target
+            outputs = torch.sigmoid(logits)
+        else:  # multiple targets
+            outputs = nn.functional.softmax(logits / self.temperature)
+
+        return outputs
+
+    def forward(self, features: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        features.update({"prediction": self._forward(features["sentence_embedding"])})
```
lewtun, line 126: I'm curious about the need for separate `_forward()` and `forward()` methods - is there a reason we can't wrap the logic in a single method?

src/setfit/modeling.py
```diff
@@ -41,13 +166,93 @@ def __init__(self, model_body=None, model_head=None, multi_target_strategy=None)
...
+            criterion = self.model_head.get_loss_fn()
+            optimizer = self._prepare_optimizer(learning_rate, body_learning_rate, l2_weight)
+
+            for epoch_idx in range(num_epochs):
+                for batch in dataloader:
+                    features, labels = batch
+                    optimizer.zero_grad()
+
+                    outputs = self.model_body(features)
+                    predictions = self.model_head._forward(outputs["sentence_embedding"])
```
lewtun, line 198: Depending on the answer to my question about two `forward()` methods, it might also be worth implementing a `__call__` method in `SetFitHead` to simplify this step a bit

src/setfit/modeling.py
```diff
@@ -41,13 +166,93 @@ def __init__(self, model_body=None, model_head=None, multi_target_strategy=None)
...
+        return dataloader
+
+    def _prepare_optimizer(
+        self,
+        learning_rate: float,
+        body_learning_rate: Optional[float],
+        l2_weight: float,
+    ):
+        body_learning_rate = body_learning_rate or learning_rate
+        optimizer = torch.optim.SGD(
```
lewtun, line 229: Would it make sense to use `AdamW` here or did you find `SGD` works better?

src/setfit/trainer.py
```diff
@@ -195,11 +197,57 @@ def call_model_init(self, params: Dict[str, Any] = None):
 
         return model
 
-    def train(self, trial: Union["optuna.Trial", Dict[str, Any]] = None):
+    def freeze(self):
+        """
+        Freeze SetFitModel's differentiable head.
+        Note: call this function only when using the differentiable head.
```
lewtun, line 203: Maybe we could do a small check on the `SetFitModel` and raise a `ValueError` if a user tries to execute this with the wrong model?

tests/test_modeling.py
```diff
@@ -72,3 +109,49 @@ def test_setfit_multilabel_classifier_chain_classifier_model_head():
     )
 
     assert type(model.model_head) is ClassifierChain
+
+
+def test_setfit_single_target_differentiable_head():
```
lewtun, line 114: I recently refactored the `test_trainer.py` tests into groups by using `unittest.TestCase`. This has the advantage of allowing us to download groups of models just once vs multiple times per test run. 

Would you be happy to group your tests in a similar way? 

(I realise the existing tests in `test_modeling.py` also need to be grouped - I'll do that later :))

lewtun (commented):
Absolutely incredible work on implementing a pure `torch` model @blakechi 🔥 !

I've left a few comments / questions, but overall this is looking really good.

Would you mind sharing a code snippet in the PR description so that others can understand how the new API should work? 

I'm also curious if you tested this implementation with some of the test datasets in our paper? It would be cool to know if (a) this implementation does better than our original one and (b) there are no subtle regressions with the `sklearn` version.

src/setfit/modeling.py
blakechi: okay, I will add them in the next commit

src/setfit/modeling.py
blakechi: Thanks!

src/setfit/modeling.py
blakechi: My original intend is to separate the usages. So `forward()` is for being compatible with Sentence-Transformer, and on the other hand, `_forward()` can be used when users feed embeddings only.

But ya, I think I can merge them together and act correspondly by detecting what format users feed into it. Will fix it in the next commit!

src/setfit/modeling.py
blakechi: Ya, I think by merging `forward()` and `_forward()` together as I answered in your question, I can simplify this step.

src/setfit/modeling.py
blakechi: sorry I used SGD for simplicity. I think maybe we can decide which one to use after running experiments?

src/setfit/trainer.py
blakechi: That's a great idea. Let me add it!

tests/test_modeling.py
blakechi: sure no problem 👍

src/setfit/trainer.py
blakechi: Just added it

src/setfit/modeling.py
blakechi: Fixed!

src/setfit/modeling.py
blakechi: Fixed!

src/setfit/modeling.py
blakechi: Just added it!

blakechi:
> Absolutely incredible work on implementing a pure `torch` model @blakechi 🔥 !
> 
> I've left a few comments / questions, but overall this is looking really good.
> 
> Would you mind sharing a code snippet in the PR description so that others can understand how the new API should work?
> 
> I'm also curious if you tested this implementation with some of the test datasets in our paper? It would be cool to know if (a) this implementation does better than our original one and (b) there are no subtle regressions with the `sklearn` version.

Thanks! Glad you like it!

Sure, will provide a snippet in the next comment. :)

Sorry I might rush a bit. I wanted to share with you the implementation to make sure the APIs are correct, so I only tested the head by running one step (forward and backward) to check its gradients. 
I will test this implementation on the test datasets. If you have some suggested scripts for me to run, I happy to know!

blakechi:
Here is the snippet for using the differentiable head (partially copied from `README.md`):
```python

num_classes = 2
train_dataset = dataset["train"].shuffle(seed=42).select(range(8 * num_classes))

# Initialize `SetFitModel` with the differentiable head
    "sentence-transformers/paraphrase-albert-small-v2",
    use_differentiable_head=True,
    head_params={"out_features": num_classes},
)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20, # The number of text pairs to generate for contrastive learning
    num_epochs=1, # The number of epochs to use for constrastive learning
    column_mapping={"sentence": "text", "label": "label"} # Map dataset columns to text/label expected by trainer
)

# Freeze head
trainer.freeze()

# Do contrastive training
trainer.train(num_epochs=1)

# Unfreeze head
trainer.unfreeze()

# Unfreeze head and freeze body
# trainer.unfreeze(keep_body_frozen=True)

# Train end-to-end
trainer.train(num_epochs=1)
```

tests/test_modeling.py
blakechi: done!

src/setfit/modeling.py
lewtun: Sounds good to me :)

blakechi:
Hi @lewtun,

Sorry for the late. I was packed by other staffs.

Ya, that sounds good to me. I ran the experiments on the dev set using `pytorch` head and they are similar to the performance on the paper. Below is the results:

| Head | SST-5 | Amazon-CF | CR | Emotion | EnronSpam | AGNews |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| pytorch | 44.4 (2.9) | 39.9 (11.7) | 89.0 (1.0) | 48.4 (5.2) | 89.1 (4.1) | 83.3 (2.9) |

I’m running experiments using `sklearn` head and will report it later once it finishes.

Sure, just resolved it!

src/setfit/modeling.py
blakechi: I will report the results using SGD later, now the first table (please check below) uses AdamW :)

blakechi: Reported the results using `SGD`, `Adam`, and `AdamW`, please see below :)

scripts/setfit/cal_score.py
```diff
@@ -0,0 +1,50 @@
+import argparse
```
lewtun, line 1: I'm curious what you need this script for? We have a `create_summary_table.py` [script](https://github.com/huggingface/setfit/blob/main/scripts/create_summary_table.py) for aggregating scores across runs, but maybe you needed something else?

scripts/setfit/run_fewshot.py
```diff
@@ -121,7 +129,19 @@ def main():
...
             )
-            trainer.train()
+            if args.classifier == "pytorch":
+                trainer.freeze()
+                trainer.train()
+                trainer.unfreeze(keep_body_frozen=args.keep_body_frozen)
+                trainer.train(
+                    num_epochs=25,
+                    body_learning_rate=1e-5,
+                    learning_rate=args.lr,  # recommand: 1e-2
```
lewtun, line 139: turbo nit:
```suggestion
                    learning_rate=args.lr,  # recommend: 1e-2
```

src/setfit/modeling.py
```diff
@@ -13,17 +13,77 @@
...
+
 logging.set_verbosity_info()
 logger = logging.get_logger(__name__)
 
-
+TokenizerOutput = Dict[str, List[int]]
 MODEL_HEAD_NAME = "model_head.pkl"
 
 
+class SetFitDataset(Dataset):
```
lewtun, line 35: I think this would belong better in `setfit.data` - WDYT?

Could we also add a one-line docstring to summarise what it's for?

src/setfit/modeling.py
```diff
@@ -34,52 +94,281 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+                The embeddings from the encoder. If using `dict` format,
+                make sure to store embeddings under the key: 'sentence_embedding'
+                and the outputs will be under the key: 'prediction'.
+            temperature (`float`, *optional*):
+                A logits' scaling factor when using multi-targets (i.e., number of targets more than 1).
+                Will override the temerature given during initialization.
+        Returns:
+        [`Dict[str, torch.Tensor]` or `torch.Tensor`]
+        """
+        is_dict = False  # whether `features` is dict or not
```
lewtun, line 162: nit: I think it helps to be explicit whenever possible 
```suggestion
        is_features_dict = False  # whether `features` is dict or not
```

src/setfit/modeling.py
```diff
@@ -34,52 +94,281 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+        if isinstance(features, dict):
+            assert "sentence_embedding" in features
+            is_dict = True
```
lewtun, line 165:
```suggestion
            is_features_dict = True
```

src/setfit/modeling.py
```diff
@@ -34,52 +94,281 @@ def __init__(self, model, max_seq_length: int, add_normalization_layer: bool) ->
...
+
+        x = features["sentence_embedding"] if is_dict else features
+        logits = self.linear(x)
+        if self.out_features == 1:  # only has one target
+            outputs = torch.sigmoid(logits)
+        else:  # multiple targets
+            temperature = temperature or self.temperature
+            outputs = nn.functional.softmax(logits / temperature, dim=-1)
+
+        if is_dict:
```
lewtun, line 175:
```suggestion
        if is_features_dict:
```

lewtun (commented):
Thanks so much for sharing the experimental results with the new API - incredible work 🔥 !

This PR is looking really close to being done - I've just left a few small nits and then we can merge 🥳 

I'm excited to see this in action!

src/setfit/modeling.py
blakechi: That's a good idea! Will also add doctoring as well

scripts/setfit/cal_score.py
blakechi: oh haha, I didn't see that script 😂
ok, I will remove `cal_score.py` since it's duplicated to `create_summary_table.py`, thanks!

src/setfit/modeling.py
blakechi: done!

scripts/setfit/cal_score.py
blakechi: done!

lewtun (approved):
Thanks for the final iteration - this looks great so I'm going to merge it now 🔥 !

Amazing contribution and thank you for working on it @blakechi 🤗

PhilipMay:
Good job @blakechi ! Many thanks.

blakechi:
> Thanks for the final iteration - this looks great so I'm going to merge it now 🔥 !
> 
> Amazing contribution and thank you for working on it @blakechi 🤗

I want to thank you for your advice and review as well, those make this PR wonderful! Really like this collaboration! 🤗 @lewtun

Issue #94, opened 2022-10-11
HFValidationError when loading the model


jrivd:
Hi there! I am trying to load a model I have stored at Google Drive for inferencing:

```
# Load SetFit model
tuned_model = SetFitModel.from_pretrained("/content/drive/My Drive/models/tuned-model")
# Run inference
tuned_model(["i didnt feel humiliated", "i feel romantic too", "im grabbing a minute to post i feel greedy wrong"])
```

But I get the following error:

`
HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/content/drive/My Drive/models /tuned-model'. Use repo_type argument if needed.
`

it works however when I load it from the same script in which I have saved it, by:

```
# Save trained model to disk
trainer.model.save_pretrained("/content/drive/My Drive/models/tuned-model")
```

What can be the problem? Can't just I save/load from pretrained to Google Drive?

Many thanks in advance for your support and terrific work.

lewtun:
Hey @jrivd can you provide a code snippet / Colab that reproduces the error? This will help debug what exactly is going on :)

jrivd:
Hi @lewtun, thanks for your response. You can see the error in action here: https://colab.research.google.com/drive/10t9QmQEe7BHIQQ8XUmw1B37vPFRfORDK?usp=sharing
Many thanks!

pdhall99:
I get the same error. When I try to load a locally saved model:
```python
from setfit import SetFitModel

model = SetFitModel.from_pretrained("/path/to/model-directory", local_files_only=True)
```
I get
```python
HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/path/to/model-directory'. Use `repo_type` argument if needed.
```

I think this could be solved by changing [these lines](https://github.com/huggingface/setfit/blob/a06be0efae1f71357aa357af8df42175f6adc91e/src/setfit/modeling.py#L84) from
```python
        if os.path.isdir(model_id) and MODEL_HEAD_NAME in os.listdir(model_id):
            model_head_file = os.path.join(model_id, MODEL_HEAD_NAME)
```
to something like
```python
        if os.path.isdir(model_id):
            if MODEL_HEAD_NAME in os.listdir(model_id):
                model_head_file = os.path.join(model_id, MODEL_HEAD_NAME)
            else:
                model_head_file = None
```

Mouhanedg56:
[...]

PR #114, opened 2022-10-19
fix: allow load of pretrained model without head

pdhall99:
Fix the issue discussed in issue https://github.com/huggingface/setfit/issues/94, allowing the load of a pretrained model without a head.

### DIFF
--- a/tests/test_modeling.py
+++ b/tests/test_modeling.py
@@ -5,7 +5,7 @@ from sklearn.multiclass import OneVsRestClassifier
 from sklearn.multioutput import ClassifierChain, MultiOutputClassifier
 
 from setfit import SetFitModel
-from setfit.modeling import sentence_pairs_generation, sentence_pairs_generation_multilabel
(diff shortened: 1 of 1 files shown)


lewtun (commented):
Thanks a lot for this fix @pdhall99 🔥 !

Would you mind adding a small unit test that tests we can load local models with these kind of directory structures? The simplest thing would be to have a test that loads a `SetFitModel` from the Hub, saves it to disk with `save_pretrained("some folder structure to be tested") and then checks we can load it with `from_pretrained()`

pdhall99:
> Would you mind adding a small unit test that tests we can load local models with these kind of directory structures?

Thanks @lewtun. I've added a couple of tests along those lines. Let me know if it's not what you had in mind.

lewtun (approved):
Thanks for adding the tests @pdhall99 🤗 ! This all LGTM so merging now :)

PR #125, opened 2022-10-26
Add support to kwargs in `compute()` method called by `trainer.evaluate()`

mpangrazzi:
Hi!

This small change makes one able to pass additional arguments to `compute()` method called by `trainer.evaluate()`.
It's needed to e.g. computing correctly `f1` metric in multilabel mode. 

Currently this does not work and it exists with error: 

```
ValueError: Target is multiclass but average='binary'. Please choose another average setting, one of [None, 'micro', 'macro', 'weighted']
```

because at least an `average="micro"` additional param is needed in `compute()` method.

Example usage after this change:

```python
# ...

model = SetFitModel.from_pretrained("sentence-transformers/all-roberta-large-v1")

trainer = SetFitTrainer(
    model=model,
    train_dataset=multilabel_train_ds,
    eval_dataset=multilabel_test_ds,
    metric="f1",
    batch_size=8,
    num_iterations=100,
    num_epochs=2,
)

trainer.train()

metrics = trainer.evaluate(average="micro")
```

Don't know if you like this way or if you want instead to have a `metric_kwargs` param when creating `SetFitTrainer`.

### DIFF
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -158,7 +158,7 @@ class SetFitTrainerTest(TestCase):
 
         trainer.train()
 
-        with self.assertRaises(ValueError):
+        with self.assertRaises(FileNotFoundError):
             trainer.evaluate()
 
     def test_trainer_raises_error_with_wrong_warmup_proportion(self):


bengsoon:
Haha @mpangrazzi You beat me to this! I was literally facing the same issue and was just about to create a pull request just exactly like this.

lewtun (commented):
Thanks a lot for catching this bug and opening a PR @mpangrazzi ! 

I'm partially in favour of having `metric_kwargs` in the trainer init for the sake of grouping the logic together in one place. WDYT?

One advantage of your approach is that it allows the user to compute multiple metrics like `micro` and `macro` without having to train the model again. Having said that, these models are super fast to train, so if users want to report multiple metrics we should probably consider refactoring things to support a `compute_metrics()` callable like the `transformers.Trainer` does :)

src/setfit/trainer.py
```diff
@@ -285,22 +285,31 @@ def train(self, trial: Union["optuna.Trial", Dict[str, Any]] = None):
         # Train the final classifier
         self.model.fit(x_train, y_train)
 
-    def evaluate(self):
-        """Computes the metrics for a given classifier."""
+    def evaluate(self, **kwargs):
```
PhilipMay, line 288: What about renaming `**kwargs` to `**evaluate_kwargs`? 
Is more self-explanatory.

lewtun:
Hi @mpangrazzi @PhilipMay I've been mulling over this feature and think the most elegant solution would be to allow `metric` to accept a callable where the user can define custom logic. This is similar to how the evaluators in `evaluate` work ([docs](https://huggingface.co/docs/evaluate/package_reference/evaluator_classes#evaluate.TextClassificationEvaluator.compute.metric)) and should cover most use cases.

If that makes sense to you, would you like to adapt this PR so that we extend the type of `metric` from `str` to `Union[str, Callable]` and have a check in `evaluate()` for which type of input we're dealing with?

mpangrazzi:
@PhilipMay @lewtun Hi! Yes, I actually like the `transformers.Trainer` implementation, especially for computing multiple metrics at once.

I can adapt this PR to accept also a `Callable` for `metric` (so it will work similar to `compute_metrics` in `transformers.Trainer`).

We can also support a `metric_kwargs` param which will be used when `metric` is a `str`, so one can avoid to define a custom function when only an additional param is needed (e.g. for multilabel `f1`). WDYT?

lewtun:
Sounds good @mpangrazzi ! Maybe we can start with just supporting a `Callable` and wait a bit before adding a `metric_kwargs` parameter. 

I'm a bit reluctant to add it right now because I know `evaluate` is planning to change the way these args are handled and we can revisit this they've released the new API

mpangrazzi:
I've added the support for a `Callable` to `metric` argument. So now it's possible to do something like the following (see also updated tests):

```python
# ...

multilabel_f1_metric = evaluate.load("f1", "multilabel")
multilabel_accuracy_metric = evaluate.load("accuracy", "multilabel")

def compute_metrics(y_pred, y_test):
    return {
        "f1": multilabel_f1_metric.compute(predictions=y_pred, references=y_test, average="micro")["f1"],
        "accuracy": multilabel_accuracy_metric.compute(predictions=y_pred, references=y_test)["accuracy"],
    }

trainer = SetFitTrainer(
    model=self.model,
    train_dataset=dataset,
    eval_dataset=dataset,
    metric=compute_metrics,
    num_iterations=self.num_iterations,
    column_mapping={"text_new": "text", "label_new": "label"},
)

trainer.evaluate() # returns a dict with "f1" and "accuracy"
```

However, it's still possible to use a `str` for metric (e.g. `f1`) and call `evaluate()` with `evaluate_kwargs` like `trainer.evaluate(average="micro")`.

Do you want to remove the latter approach or can we keep both?

src/setfit/trainer.py
```diff
@@ -372,22 +373,37 @@ def train(
                 show_progress_bar=True,
             )
 
-    def evaluate(self):
-        """Computes the metrics for a given classifier."""
+    def evaluate(self, **evaluate_kwargs):
```
PhilipMay, line 376: IMO this would be better like this:

`    def evaluate(self, evaluate_kwargs):`

So without the `**`. The reason is that if there are the `**` then evaluate_kwargs cates 
all other function params. What if we want to extend the evaluate function to also pass 
args to someting else in the function? How do we then add an other param for that?

PhilipMay: Or just remove the metric_kwargs as suggested by @lewtun

src/setfit/trainer.py
```diff
@@ -69,7 +70,7 @@ def __init__(
         train_dataset: "Dataset" = None,
         eval_dataset: "Dataset" = None,
         model_init: Callable[[], "SetFitModel"] = None,
-        metric: str = "accuracy",
+        metric: Union[str, Callable] = "accuracy",
```
PhilipMay, line 73: Would it be possible to be more specific with the types that Callable accepts?

see https://docs.python.org/3/library/typing.html#callable

src/setfit/trainer.py
```diff
@@ -372,22 +373,37 @@ def train(
...
         y_pred = self.model.predict(x_test)
 
-        return metric_fn.compute(predictions=y_pred, references=y_test)
+        if isinstance(self.metric, str):
+            metric_config = "multilabel" if self.model.multi_target_strategy is not None else None
+            metric_fn = evaluate.load(self.metric, config_name=metric_config)
+
+            return metric_fn.compute(predictions=y_pred, references=y_test, **evaluate_kwargs)
+
+        if callable(self.metric):
```
PhilipMay, line 403: This should be an `elif`.

PhilipMay, line 403: Should this better be an `elif`?

```
        elif callable(self.metric):
```

src/setfit/trainer.py
mpangrazzi: Thanks for the references! Maybe it's better to simply remove kwargs and leave the callable option, to avoid doing two different things at once in this PR

src/setfit/trainer.py
mpangrazzi: Good point, sure

src/setfit/trainer.py
lewtun: Yes, I agree with removing kwargs in this PR - I'd like to revisit this once `evaluate` has implemented a new API for how one defines metrics with runtime args (like `f1` score). These will soon be defined entirely by `evaluate.load()` and that should make it easy to pass everything at once to the `metric` arg

src/setfit/trainer.py
lewtun: Yes, then we could end the logic with an `else` to raise the error :)

lewtun (commented):
Thanks a lot for refactoring the API @mpangrazzi !

This is looking great and if we can remove the `evaluate_kwargs` for now, I think this will be good to merge 🔥

src/setfit/trainer.py
PhilipMay: Yes. That was also my first idea. But if we do that we could maybe better do it in the `__init__` function - because there
`self.metric` is set.

src/setfit/trainer.py
```diff
@@ -39,8 +39,9 @@ class SetFitTrainer:
         model_init (`Callable[[], SetFitModel]`, *optional*):
             A function that instantiates the model to be used. If provided, each call to [`~SetFitTrainer.train`] will start
             from a new instance of the model as given by this function when a `trial` is passed.
-        metric (`str`, *optional*, defaults to `"accuracy"`):
-            The metric to use for evaluation.
+        metric (`str` or `Callable`, *optional*, defaults to `"accuracy"`):
+            The metric to use for evaluation. If a string is provided, it must be a key present in the evaluation dataset.
```
lewtun, line 43: I wasn't sure what you meant by "key", but maybe this is a bit clearer?
```suggestion
            The metric to use for evaluation. If a string is provided, we treat it as the metric name and load it with default settings.
```

tests/test_trainer.py
```diff
@@ -107,6 +108,59 @@ def test_column_mapping_multilabel(self):
...
+    def test_raise_when_metric_value_is_invalid(self):
+        dataset = Dataset.from_dict(
+            {"text_new": ["a", "b", "c"], "label_new": [0, 1, 2], "extra_column": ["d", "e", "f"]}
+        )
+
+        trainer = SetFitTrainer(
+            model=self.model,
+            train_dataset=dataset,
+            eval_dataset=dataset,
+            metric=42,  # invalid metric value
```
lewtun, line 154: How about we try a `str` with a metric that almost surely won't exist in `evaluate`?
```suggestion
            metric="this-metric-does-not-exist",  # invalid metric value
```

lewtun (approved):
Great work @mpangrazzi - this is very clean now 🤗 !

I've left a few nits and then we can merge 🥳

src/setfit/trainer.py
mpangrazzi: yes sorry I was referring to metric name 😅

tests/test_trainer.py
mpangrazzi: sure!

lewtun (approved):
Thanks for iterating! Looks great now :)

Issue #129, opened 2022-10-28
Use amp / FP16


PhilipMay:
Hi, since the sentence transformers support FP16 training (by setting `use_amp=True`)
can this also be build in here?

Not sure if this is a FR or a question. :-)

PhilipMay:
Should be easy to add it here: 

https://github.com/huggingface/setfit/blob/7735e8e3b208edb8dfb549beb16e585453c5f44e/src/setfit/trainer.py#L276

PhilipMay:
see PR #134

PhilipMay:
closing this because PR is merged

PR #134, opened 2022-10-30
add option to use amp / FP16

PhilipMay:
PR to solve #129


lewtun (approved):
Thanks for the nice addition!

PR #139, opened 2022-10-31
add num_epochs to train_step calculation

### LINKED ISSUE #137: Warmup steps calculation should not ignore epochs.
The warmup steps calculation should not ignore epochs. See here:

https://github.com/huggingface/setfit/blob/7735e8e3b208edb8dfb549beb16e585453c5f44e/src/setfit/trainer.py#L275

The bug is where train steps is calculates here:

https://github.com/huggingface/setfit/blob/7735e8e3b208edb8dfb549beb16e585453c5f44e/src/setfit/trainer.py#L267

the right implementation is here (but not always executed):

https://github.com/huggingface/setfit/blob/7735e8e3b208edb8dfb549beb16e585453c5f44e/src/setfit/trainer.py#L253

I can provide a PR.

PhilipMay:
fixes #137

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -264,7 +264,7 @@ class SetFitTrainer:
 
             train_dataloader = DataLoader(train_examples, shuffle=True, batch_size=self.batch_size)
             train_loss = self.loss_class(self.model.model_body)
-            train_steps = len(train_dataloader)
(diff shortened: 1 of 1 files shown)


lewtun (approved):
Nice catch, thanks for the fix!

PR #148, opened 2022-11-03
fix: templated examples copy empty vector

pdhall99:
Fix bug with `data.add_templated_examples`, raised by @attekei.


lewtun (approved):
Thanks for the fix!

PR #150, opened 2022-11-04
Small fix on hyperparameter search

Mouhanedg56:
This fix an issue when initialising the model with head parameters: we should use `clf` object rather than creating a new logistic regression instance with no parameter when `multi_target_strategy` is not defined.


lewtun (commented):
Thanks for the fix!

Would it make sense to have a small unit test that catches this?

Mouhanedg56:
Sure. unit test added for the fix ✔️

lewtun (approved):
Thanks for adding the test!

PR #155, opened 2022-11-06
Fix typo: `temerature` => `temperature`

tomaarsen:
Hello!

This PR speaks for itself.
This is very promising work, well done all.

- Tom Aarsen


lewtun (approved):
Thanks for the fix and kind words @tomaarsen !

Issue #78, opened 2022-10-02
Onnx support?


blakechi:
Hi, really like this work!

Given its advantage on faster inference, have you considered adding support functions, like the example below, to compile `SetFitTrainer` into the onnx format for production-wise usage?

If that sounds promising, I will be happy to make this feature work!

Example:
```python
# Train 
trainer.train()

# Compile to onnx
onnx_path = "path/to/store/compiled/model.onnx"
trainer.to_onnx(onnx_path, **onnx_related_kwargs)
```

lewtun:
Hi @blakechi thanks for your interest in our work! Instead of compiling `SetFitTrainer` into ONNX, I think it would be better to have something like an `onnx_export()` function that lives inside a `setfit.onnx` module.

Ideally, this function would take a `SetFitModel` as input and output `model.onnx`, similar to what was done for Stable Diffusion here: https://github.com/huggingface/diffusers/blob/main/scripts/convert_stable_diffusion_checkpoint_to_onnx.py#L32

Since the current implementation of `SetFitModel` uses `scikit-learn` estimators for the classification head, it might be best to wait until we implement a pure PyTorch version in #8

blakechi:
Hi @lewtun, 
Thanks for your reply! `onnx_export` for `SetFitModel` sounds good to me.

Sure, happy to lend a hand on onnx after #8. :grinning:

nbertagnolli:
I'm not sure if this is helpful, but I was working on deploying some of these models using ONNX and this is what I came up with so far.  If others are looking for a place to start here is some code that will convert the base model and the head and then you can run them separately. I haven't been able to merge them into one graph yet but hopefully it's a start while we wait for #8 :).

`
from skl2onnx import convert_sklearn
from skl2onnx.common.data_types import FloatTensorType
from onnxruntime import InferenceSession
from pathlib import Path
from setfit import SetFitModel
from transformers import AutoTokenizer
from transformers.convert_graph_to_onnx import convert
import torch
import numpy as np
import sys


def mean_pooling_np(model_output: np.array, attention_mask: np.array):
    token_embeddings = model_output[0] 
    input_mask_expanded = np.broadcast_to(
        np.expand_dims(attention_mask, axis=2), token_embeddings.shape
    )
    sum_embeddings = np.sum(input_mask_expanded * token_embeddings, axis=1)
    sum_mask = np.clip(input_mask_expanded.sum(1), 1e-9, sys.maxint)
[...]

PR #156, opened 2022-11-07
Added initial onnx export function

nbertagnolli:
This PR is an initial attempt to address #78.  It creates a function called `export_onnx` which currently handles models with a sklearn based head. It will convert the body and the head to onnx format and merge them with a custom mean_pooling onnx layer in between.  I wanted to open this to get the conversation started I'd love feedback and hope to make this something generally useful : ).

Some notes:
1. I couldn't find an implementation of the attention based mean pooling anywhere so I made one. If there is already an implementation that would probably be better than what I put together.
2. I'm not sure how you want to handle the onnx dependencies now @lewtun Should we put them as a required pacakge to install? Should we do the imports in the export function and have the user custom install them if they need?
3. I owe you some tests 😅


blakechi:
Hi @nbertagnolli,

Thanks for creating the first attempt! For your first note, I think we can export the mean pooling layer together with the text encoder. I created a [notebook](https://drive.google.com/file/d/1KvN95i6vZlBW-eecq_waQiCwjO3vPvFR/view?usp=share_link) on my laptop to demo it. And I think by doing so, we can bypass the `transformers.convert_graph_to_onnx` issue you mentioned in your code.

Another thing is since we both need `model` and `pooler` in the notebook, how about we extract those two before [deciding `pytorch` or `sklearn` head](https://github.com/huggingface/setfit/pull/156/files#diff-32f86d0e3df2a295f2a7c5f6affc1626407031fc99e05c1af4d9423cf74a64b7R88)?

Please let me know what you think :)

src/setfit/onnx.py
```diff
@@ -0,0 +1,347 @@
...
+            sklearn.estimator the models will be saved in
+        use_external_format: Split the model definition from its parameters to allow
+            model bigger than 2GB
+    Returns:
+    """
+
+    model = SetFitModel.from_pretrained(model_name)
+
+    # Check to see if the model uses a sklearn head or a torch dense layer.
+    if issubclass(type(model.model_head), models.Dense):
```
blakechi, line 88: How about before this if-else statement, we extract `model` and `pooler` as mentioned in this [notebook](https://drive.google.com/file/d/1KvN95i6vZlBW-eecq_waQiCwjO3vPvFR/view?usp=share_link)? So you can continue on connecting those two with the `sklearn` head and I can do so on the `pytorch` head @nbertagnolli

nbertagnolli: This is awesome! Yeah pulling the pooler out and attaching them like this is way better.   I'll update the PR. Thanks! 
🙏 @blakechi 🙏

blakechi: Glad you like it!
ok, I will add my park as well after your update

nbertagnolli: I added in a bunch of your suggestions : ). Thanks for all the great input! Let me know if there is anything else you see that could be better.

src/setfit/onnx.py
```diff
@@ -0,0 +1,177 @@
...
+            sklearn.estimator the models will be saved in
+        ignore_ir_version: Whether to ignore the IR version used in sklearn. The version is often missmatched
+            with the transformer models. Setting this to true coerces the versions to be the same. This might
+            cause errors but in practice works.  If this is set to False you need to ensure that the IR versions
+            align between the transformer and the sklearn onnx representation.
+    Returns:
+    """
+
+    # Load the model and get all of the parts.
+    model = SetFitModel.from_pretrained(model_name)
```
kgourgou, line 104: Looks really nice!

Recently I've recently been using more 

```
SetFitModel(body=SomeSentenceTransformer, head=SomeHead)
``` 

instead of "from_pretrained()". The advantage is that if I implement a model head in PyTorch, then I can easily spin a SetFit model with my custom head. I wonder if something like that makes sense here?

nbertagnolli: That makes sense! I'll update this afternoon : )

src/setfit/onnx.py
blakechi: Added my part at here: nbertagnolli/setfit#1

src/setfit/onnx.py
kgourgou: Thank you, it looks good!

src/setfit/onnx.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns:
+    """
+
+    # Load the model and get all of the parts.
+    transformer = model_body_sentence_transformer._modules["0"]
+    tokenizer = transformer.tokenizer
+    max_length = transformer.max_seq_length
+    model_body = transformer.auto_model
+    model_pooler = model_body_sentence_transformer._modules["1"]
+    model_head = model.model_head
```
kgourgou, line 115: Just highlighting this line; with the new commits here, there is no `model` variable anymore.

blakechi:
Hi @nbertagnolli , it seems like we have implemented the basic functionalities for exporting both kinds of heads. Would you like to refactor the `export_onnx` function such that some common snippets can be pulled out of the `if-else` statement to prevent duplication? If not, I would be more than happy to do it. 😄 So no pressure.

I find some parts we can improve as below, if you have others, feel free to share it as well

- `onnx.save`([here](https://github.com/huggingface/setfit/pull/156/files#diff-32f86d0e3df2a295f2a7c5f6affc1626407031fc99e05c1af4d9423cf74a64b7R174), [here](https://github.com/huggingface/setfit/pull/156/files#diff-32f86d0e3df2a295f2a7c5f6affc1626407031fc99e05c1af4d9423cf74a64b7R227)) can be pulled out of `if-else`.
- I suggest we add the tokenizer's metadata in the exported ONNX file for both heads such that users have a clue on which tokenizer they should get and how to use it (the [ONNX runtime API](https://github.com/huggingface/setfit/issues/78#issuecomment-1304593409) can also benefit from it)
- Wrap this [snippet](https://github.com/huggingface/setfit/pull/156/files#diff-32f86d0e3df2a295f2a7c5f6affc1626407031fc99e05c1af4d9423cf74a64b7R190-R205) into a function that accept necessary arguments

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+from typing import Optional, Union
+import torch
+from sklearn.linear_model import LogisticRegression
+
+
+class ONNXSetFitModel(torch.nn.Module):
+    def __init__(
+        self,
+        model_body,
+        pooler,
```
lewtun, line 15: I guess this can be a generic `Callable` right?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def forward(self, input_ids, attention_mask, token_type_ids):
+        hidden = self.model_body(input_ids, attention_mask, token_type_ids)
+        hidden = {"token_embeddings": hidden[0], "attention_mask": attention_mask}
+
+        hidden = self.pooler(hidden)["sentence_embedding"]
+        if self.model_head is None:
+            return hidden
+
+        out = self.model_head(hidden)
```
lewtun, line 32: Would this be more descriptive or does it only apply to the PyTorch model?
```suggestion
        logits = self.model_head(hidden)
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        self.model_body = model_body
+        self.pooler = pooler
+        self.model_head = model_head
+
+    def forward(self, input_ids, attention_mask, token_type_ids):
+        hidden = self.model_body(input_ids, attention_mask, token_type_ids)
+        hidden = {"token_embeddings": hidden[0], "attention_mask": attention_mask}
+
+        hidden = self.pooler(hidden)["sentence_embedding"]
+        if self.model_head is None:
```
lewtun, line 29: Should we allow a user to export a model without a head? Just wondering whether to raise a `ValueError` here

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self.model_body = model_body
+        self.pooler = pooler
+        self.model_head = model_head
+
+    def forward(self, input_ids, attention_mask, token_type_ids):
+        hidden = self.model_body(input_ids, attention_mask, token_type_ids)
+        hidden = {"token_embeddings": hidden[0], "attention_mask": attention_mask}
+
+        hidden = self.pooler(hidden)["sentence_embedding"]
```
lewtun, line 28: nit:
```suggestion
        embeddings = self.pooler(hidden)["sentence_embedding"]
```

Also is the `sentence_embedding` key necessary? I was thinking we could just require people to return a `torch.Tensor`

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        model_head: Optional[Union[torch.nn.Module, LogisticRegression]] = None,
+    ):
+        super().__init__()
+
+        self.model_body = model_body
+        self.pooler = pooler
+        self.model_head = model_head
+
+    def forward(self, input_ids, attention_mask, token_type_ids):
+        hidden = self.model_body(input_ids, attention_mask, token_type_ids)
```
lewtun, line 25: nit:
```suggestion
        hidden_states = self.model_body(input_ids, attention_mask, token_type_ids)
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+import warnings
+from typing import Optional, Union
+import torch
+from sklearn.linear_model import LogisticRegression
+
+
+class ONNXSetFitModel(torch.nn.Module):
+    def __init__(
+        self,
+        model_body,
```
lewtun, line 14: could we have some types here please :)

src/setfit/onnx.py
lewtun, line 15: What do you think about defining a function like this:

```python
def mean_pooling(token_embeddings, attention_mask) -> torch.Tensor:
    input_mask_expanded = attention_mask.unsqueeze(-1).expand(token_embeddings.size()).float()
    return torch.sum(token_embeddings * input_mask_expanded, 1) / torch.clamp(input_mask_expanded.sum(1), min=1e-9)
```

and using it as the default for the `pooler`? I took this from the model card of `sentence-transformers`: https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2#usage-huggingface-transformers

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        pooler,
+        model_head: Optional[Union[torch.nn.Module, LogisticRegression]] = None,
+    ):
+        super().__init__()
+
+        self.model_body = model_body
+        self.pooler = pooler
+        self.model_head = model_head
+
+    def forward(self, input_ids, attention_mask, token_type_ids):
```
lewtun, line 24: Can we have some types here too please?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+        hidden = self.pooler(hidden)["sentence_embedding"]
+        if self.model_head is None:
+            return hidden
+
+        out = self.model_head(hidden)
+        return out
+
+
+def sklearn_head_to_onnx(model_head: LogisticRegression, opset: int):
```
lewtun, line 36: nit:
```suggestion
def export_sklearn_head_to_onnx(model_head: LogisticRegression, opset: int):
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Convert the sklearn head from a SetFitModel to ONNX format.
```
lewtun, line 38: turbo nit:
```suggestion
    Convert the Scikit-Learn head from a SetFitModel to ONNX format.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+    :param model_head: The trained SetFit model_head.
```
lewtun, line 40: Would you mind aligning the docstring format with those in the code base?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+    # Check if skl2onnx is installed
+    try:
+        from skl2onnx.common.data_types import guess_data_type
+        from skl2onnx import convert_sklearn
+        import onnxconverter_common
+        from skl2onnx.sklapi import CastTransformer
+        from sklearn.pipeline import Pipeline
+    except ImportError:
+        msg = (
+            "skl2onnx must be installed in order to convert a model with an sklearn head."
```
lewtun, line 58: Can we provide the `pip` install command here to help the users?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+        raise ImportError(msg)
+
+    # Check to see that the head has a coef_
```
lewtun, line 62: should we also check if the head has been fitted or is that captured by this check as well?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Determine the initial type and the shape of the output.
+    input_shape = (None, *model_head.coef_.shape[1:])
+    dtype = guess_data_type(model_head.coef_, shape=input_shape)[0][1]
+    dtype.shape = input_shape
+
+    # If the datatype of the model is double we need to cast the outputs
+    # from the setfit model to doubles for compatibility inside of ONNX.
+    if type(dtype) == onnxconverter_common.data_types.DoubleTensorType:
+        # TODO:: TALK ABOUT FLOAT CONVERSION ISSUES.
```
lewtun, line 76: Can you share what this refers to?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+def export_onnx(
+    model_body_sentence_transformer,
+    model_head,
+    opset: int,
+    output: str,
+    ignore_ir_version: bool = True,
+):
+    """
+    Export a PyTorch backed setfit model to ONNX Intermediate Representation.
```
lewtun, line 103:
```suggestion
    Export a PyTorch backed SetFit model to ONNX Intermediate Representation.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Args:
+        model_body_sentence_transformer: The model_body from a SetFitModel.model_body. This should be a
+            SentenceTransformer.
```
lewtun, line 107: I think we can indicate this in the typing right?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        model_head: The model_head from SetFitModel.model_head. This can be either a dense
```
lewtun, line 108: same comment as above:
```suggestion
        model_head: The SetFit model head. This can be either a dense
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+    model_body_sentence_transformer,
+    model_head,
+    opset: int,
+    output: str,
+    ignore_ir_version: bool = True,
+):
+    """
+    Export a PyTorch backed setfit model to ONNX Intermediate Representation.
+
+    Args:
```
lewtun, line 105: Could we add some types to these args please?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+        model_body_sentence_transformer: The model_body from a SetFitModel.model_body. This should be a
+            SentenceTransformer.
+        model_head: The model_head from SetFitModel.model_head. This can be either a dense
+            layer SetFitHead or a Sklearn estimator.
+        opset: The actual version of the ONNX operator set to use.  The final opset used
+            might be lower. ONNX will use the highest version supported by both the
+            sklearn head and the model body. If versions can't be rectified an error
+            will be thrown.
+        output: The path where will be stored the generated ONNX model. If the head is an
```
lewtun, line 114: nit:
```suggestion
        output_path: The path where will be stored the generated ONNX model. If the head is an
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+    tokenizer_kwargs = dict(
+        max_length=max_length,
+        padding="max_length",
+        return_attention_mask=True,
+        return_token_type_ids=True,
+        return_tensors="pt",
+    )
+    dummy_sample = "It's a test."
+    dummy_inputs = tokenizer(dummy_sample, **tokenizer_kwargs)
+    symbolic_names = {0: "batch_size", 1: "max_seq_len"}
```
lewtun, line 141: Maybe it's a bit clearer to do something like the following?

```python
input_names = list(dummy_inputs.keys())
output_names = ["logits"]
# dynamic axes
dynamic_axes_input = {}
for input_name in input_names:
    dynamic_axes_input[input_name] = {0: "batch_size", 1: "sequence"}

dynamic_axes_output = {}
for output_name in output_names:
    dynamic_axes_output[output_name] = {0: "batch_size", 1: "sequence"}

dynamic_axes = {**dynamic_axes_input, **dynamic_axes_output}
```

lewtun (commented):
Thanks a lot for this advanced feature @nbertagnolli - very cool work 🔥 !!

I've left a few comments, but the general API is looking quite good. Would you mind sharing a few code snippets showing how this works in practice?

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+from sentence_transformers import models
+import numpy as np
+import onnx
+import warnings
+from typing import Optional, Union
+import torch
+from sklearn.linear_model import LogisticRegression
+
+
+class ONNXSetFitModel(torch.nn.Module):
```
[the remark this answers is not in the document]
nbertagnolli: Thanks for the comments @lewtun ! I'll start working through them : ).

For the usage examples, would you like those in the README, in a notebook under notebooks, or both? : )

src/setfit/onnx.py
nbertagnolli: I believe the type of the pooler is `sentence_transformers.models.Pooling.Pooling` which is a `nn.Module`. However,  we grab the "sentence_embedding" output from the pooling layer so a generic nn.Module wouldn't work.  What are your thoughts? Should we say the type of Pooler is Pooling or nn.Module and add some caveats in the docstring?

src/setfit/onnx.py
nbertagnolli: Great suggestion! Yeah if we remove this key access then people can use other poolers more easily.

src/setfit/onnx.py
nbertagnolli: Of course! Sorry 🙇

src/setfit/onnx.py
nbertagnolli: Revisiting this I don't think it's an issue since we are going float to double. I was worried about doing a type conversion with floats and doubles.  If you used something like a random forest here and did a double to float conversion you could have really weird results since the function isn't continuous.  See this for more details. https://onnx.ai/sklearn-onnx/auto_tutorial/plot_ebegin_float_double.html

I'm going to remove this todo because since we are going from float to double we're increasing precision not decreasing it 😅.

src/setfit/onnx.py
```diff
@@ -0,0 +1,227 @@
...
+    if type(dtype) == onnxconverter_common.data_types.DoubleTensorType:
+        # TODO:: TALK ABOUT FLOAT CONVERSION ISSUES.
+        sk_model = Pipeline(
+            [("castdouble", CastTransformer(dtype=np.double)), ("head", model_head)]
+        )
+    else:
+        sk_model = model_head
+
+    # Convert sklearn head into ONNX format
+    initial_type = [("model_head", dtype)]
```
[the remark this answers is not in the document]
nbertagnolli: totes!

src/setfit/onnx.py
nbertagnolli: yessir! Done : )

src/setfit/onnx.py
nbertagnolli: Let me know if my solution addresses this comment : ).

src/setfit/onnx.py
nbertagnolli: Sorry, this is confusing.  This OnnxSetFitModel is a wrapper that makes it easy to use torch.onnx.export to export the model_body with the pooling.  In the case of the torch based head we can use this to do the whole onnx export. However when we have an sklearn head we need to export the model_body with the pooler and then the sklearn head.  This was my attempt to make it easy to do just two onnx exports but then it means that the class name doesn't fully represent what is happening here.

src/setfit/onnx.py
blakechi: I think `sentence_embedding` is necessary since now we use `sentence-transformers`'s pooling layer.
Maybe instead of remove it, adding a if-else to check whether the pooler comes from `sentence-transformers`?

nbertagnolli: I was thinking to make the type of Pooler `Optional[Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]]` Then to pass in the Pooler like we did before you could do:

```python
pooling_layer = model_body._modules["1"]
model_pooler = lambda x: pooling_layer(x)["sentence_embedding"]
OnnxSetFitModel(transformer, model_pooler, model_head).cpu()
```

This would let people pass in normal Callable's too. Thoughts?

src/setfit/onnx.py
nbertagnolli: I pulled some of this out into a function let me know if it's clearer now : ).

nbertagnolli:
Thanks for all of the great feedback everyone! I really appreciate all of the insightful comments : ). I think 🤞 I've addressed all of the comments. Please let me know if I missed anything and I'll take another crack at it.

src/setfit/onnx.py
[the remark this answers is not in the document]
lewtun: Let's start with a notebook for now as the README is getting rather long :)

I've opened an issue https://github.com/huggingface/setfit/issues/181 to have real documentation and we can add the ONNX details there once it's up

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+import numpy as np
+import onnx
+import torch
+from sentence_transformers import SentenceTransformer, models
+from sklearn.linear_model import LogisticRegression
+from torch import nn
+from transformers.modeling_utils import PreTrainedModel
+
+
+def mean_pooling(token_embeddings: torch.Tensor, attention_mask: torch.Tensor) -> torch.Tensor:
```
lewtun, line 13: Very nice!

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    """Perform attention aware mean pooling.
```
lewtun, line 14: turbo nit:
```suggestion
    """Perform attention-aware mean pooling.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+
+    This method takes in embeddings of shape (batch, sequence, embedding_size) and performs average
+    pooling across the sequence dimension to yield embeddings of size (batch, embedding_size).
+
+    From:
+    https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2#usage-huggingface-transformers
```
lewtun, line 20: Let's cite the source in case the model repo changes :)
```suggestion
    https://github.com/UKPLab/sentence-transformers/blob/0b5ef4be93d2b21de3a918a084b48aab6ba48595/sentence_transformers/model_card_templates.py#L134
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+        (`torch.Tensor`) The mean pooled embeddings of size (batch, embedding_size).
+    """
+    input_mask_expanded = attention_mask.unsqueeze(-1).expand(token_embeddings.size()).float()
+    return torch.sum(token_embeddings * input_mask_expanded, 1) / torch.clamp(input_mask_expanded.sum(1), min=1e-9)
+
+
+class OnnxSetFitModel(torch.nn.Module):
+    """A wrapper around SetFit model body, pooler, and model head which makes ONNX exporting easier.
+
+    This wrapper allows us to create a nn.Module with different levels of connectivity. We can set
```
lewtun, line 38: nit:
```suggestion
    This wrapper creates a nn.Module with different levels of connectivity. We can set
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    model_body and pooler and have a Module which maps inputs to embeddings or we can set all three
```
lewtun, line 39:
```suggestion
    `model_body` and `pooler` and have a Module which maps inputs to embeddings or we can set all three
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    and have a model which maps inputs to final predictions. This is useful because torch.onnx.export
```
lewtun, line 40:
```suggestion
    and have a model which maps inputs to final predictions. This is useful because `torch.onnx.export`
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    will work with a nn.Module.
```
lewtun, line 41:
```suggestion
    will work with a `nn.Module`.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Attributes:
+        model_body (`PreTrainedModel`): The pretrained model body of a setfit model.
+        pooler (`nn.Module` or `Callable[[torch.Tensor], torch.Tensor]`): The callable function that can
```
lewtun, line 45: I think we can use `Union` here for these types
```suggestion
        pooler (`Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]`, *optional*, defaults to `None`): The callable function that can
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+            map  tensors of shape (batch, sequence, embedding_dim) to shape (batch, embedding_dim).
+        model_head: (`torch.nn.Module` or `LogisticRegression`): The model head from the pretrained SetFit model.
```
lewtun, line 47: Use `nn.Module` for consistency with above?
```suggestion
        model_head: (`Union[nn.Module, LogisticRegression]`, *optional*, defaults to `None`): The model head from the pretrained SetFit model.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+            if it is None then the resulting OnnxSetFitModel forward will return embeddings instead of predictions.
```
lewtun, line 48:
```suggestion
            If `None`, then the resulting `OnnxSetFitModel.forward` method will return embeddings instead of predictions.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+        self,
+        model_body: PreTrainedModel,
+        pooler: Optional[Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]] = None,
+        model_head: Optional[Union[torch.nn.Module, LogisticRegression]] = None,
+    ):
+        super().__init__()
+
+        self.model_body = model_body
+        if pooler is None:
+            self.pooler = mean_pooling
```
lewtun, line 61: Maybe we can use the logger here to indicate the no pooler was provided, so we're defaulting to mean pooling?

src/setfit/onnx.py
lewtun: Thanks, this makes sense!

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+        # If head is set then we have a fully torch based model and make the final predictions
+        # with the head.
+        out = self.model_head(embeddings)
+        return out
+
+
+def export_onnx_setfit_model(setfit_model: OnnxSetFitModel, inputs, output_path, opset):
+    """Export the OnnxSetFitModel.
+
+    This exports the model created by the OnnxSetFitModel wrapper using torch.onnx.export.
```
lewtun, line 86:
```suggestion
    This exports the model created by the `OnnxSetFitModel` wrapper using `torch.onnx.export`.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+            return embeddings
+
+        # If head is set then we have a fully torch based model and make the final predictions
+        # with the head.
+        out = self.model_head(embeddings)
+        return out
+
+
+def export_onnx_setfit_model(setfit_model: OnnxSetFitModel, inputs, output_path, opset):
+    """Export the OnnxSetFitModel.
```
lewtun, line 84:
```suggestion
    """Export the `OnnxSetFitModel`.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+
+    This exports the model created by the OnnxSetFitModel wrapper using torch.onnx.export.
+
+    Args:
+        setfit_model (`OnnxSetFitModel`): The OnnxSetFitModel we want to export to .onnx format.
+        inputs (`Dict[str, torch.Tensor]`): The inputs we would hypothetically pass to the model. These are
+            generated using a tokenizer.
+        output_path (`str`): The local path to save the onnx model to.
+        opset (`int`): The ONNX opset to use for the export.
```
lewtun, line 93: What do you think about setting a good default value for `opset` since it might be annoying for users to have to look it up

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+        )
+
+
+def export_sklearn_head_to_onnx(model_head: LogisticRegression, opset: int) -> onnx.onnx_ml_pb2.ModelProto:
+    """Convert the Scikit-Learn head from a SetFitModel to ONNX format.
+
+    Args:
+        model_head (`LogisticRegression`): The trained SetFit model_head.
+        opset (`int`): The ONNX opset to use for optimizing this model. The opset is not
+            gauranteed and will default to the maximum version possible for the sklearn
```
lewtun, line 126:
```suggestion
            guaranteed and will default to the maximum version possible for the sklearn
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+            model.
+
+    Returns:
+        [`onnx.onnx_ml_pb2.ModelProto`] The ONNX model generated from the sklearn head.
+
+    Raises:
+        ImportError: If skl2onnx is not installed an error will be raised asking
```
lewtun, line 133:
```suggestion
        ImportError: If `skl2onnx` is not installed an error will be raised asking
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    try:
+        import onnxconverter_common
+        from skl2onnx import convert_sklearn
+        from skl2onnx.common.data_types import guess_data_type
+        from skl2onnx.sklapi import CastTransformer
+        from sklearn.pipeline import Pipeline
+    except ImportError:
+        msg = """
+        skl2onnx must be installed in order to convert a model with an sklearn head.
+        Please install with pip install skl2onnx.
```
lewtun, line 147:
```suggestion
        Please install with `pip install skl2onnx`.
```

src/setfit/onnx.py
```diff
@@ -0,0 +1,281 @@
...
+    # Check if skl2onnx is installed
+    try:
+        import onnxconverter_common
+        from skl2onnx import convert_sklearn
+        from skl2onnx.common.data_types import guess_data_type
+        from skl2onnx.sklapi import CastTransformer
+        from sklearn.pipeline import Pipeline
+    except ImportError:
+        msg = """
+        skl2onnx must be installed in order to convert a model with an sklearn head.
```
lewtun, line 146:
```suggestion
        `skl2onnx` must be installed in order to convert a model with an sklearn head.
```

lewtun (commented):
Thanks a lot for iterating on this feature @nbertagnolli - it's looking very nice 🚀 !

I've left some nits, but apart from that I think the only missing piece is some unit tests. You could pick a small checkpoint from the Hub and verify that the forward pass of the model in PyTorch vs ONNX Runtime is the same (within some tolerance)

src/setfit/onnx.py
nbertagnolli: That's a great suggestion! I've had 12 work well so I'll set that to start : )

nbertagnolli:
Sorry about the delay was out a bit last week.  I think I've addressed that next round of comments : ). I also added a small notebook demonstrating the usage of `export_onnx` function and a basic test to ensure that the models outout matches the original model. Please let me know if there is anything else you notice we should improve :D.

nbertagnolli:
@lewtun I'm not sure how you'd like to manage dependencies in setup.py. I added them to the `TESTS_REQUIRE` dependencies list because they are needed for testing. I don't know if that makes the most sense or if there is something better we could do.

nbertagnolli:
I found the problem : ), I wasn't getting the maximum opset for the sklearn head! Yay tests!

lewtun:
Sorry for the delay in reviewing this @nbertagnolli - on it right now!

lewtun (approved):
Thanks a lot for iterating on this killer feature @nbertagnolli - it's looking great!

I've left some tiny nits and a suggestion to move the files into a dedicated `setfit.exporters` module so that we can easily support new exporters like OpenVINO.

After that we can finally merge this beast 🔥 !

nbertagnolli:
Thanks for the great comments as always! I made the new exporters module and put onnx in there : ). Let me know if you see anything else!

tomaarsen:
I'm missing a `setfit/exporters/__init__.py` to turn `setfit.exporters` into a module. Furthermore, `setfit/exporters/utils.py` is not included, despite an attempt to import from it.

- Tom Aarsen

lewtun:
Hey @nbertagnolli thanks for the last changes, would you mind running `make style && make quality` locally and pushing the formatted code? I think that should fix the CI :)

nbertagnolli:
Sorry about all the style commits that's what i get for being lazy and not running the style checker locally 😅 😅.

lewtun:
Wonderful, thank you so much for this 🔥 !!

PR #159, opened 2022-11-07
add more loss function options

PhilipMay:
Add more options to the non default loss functions.
This is `distance_metric` and `margin`.
The default values are selected in a way that this is no breaking change.

## Todo

- [x] add docstrings


PhilipMay:
This is done and can be merged from my point of view.

lewtun (approved):
Thanks for adding these additional args @PhilipMay 🤗 !

I think at some point we'll need to refactor this out to a dedicated class for the training args (like the `transformers.Trainer`), but happy to have this included for now :)

PhilipMay:
> Thanks for adding these additional args @PhilipMay 🤗 !
> 
> I think at some point we'll need to refactor this out to a dedicated class for the training args (like the `transformers.Trainer`), but happy to have this included for now :)

Should we collect this "future ideas" anywhere?

Issue #169, opened 2022-11-10
RuntimeError: Expected all tensors to be on the same device


creatorrr:
**Code:**

```python
model = SetFitModel.from_pretrained(
      "sentence-transformers/paraphrase-mpnet-base-v2",
       use_differentiable_head=True,
       head_params={"out_features": num_classes},
    )

trainer = SetFitTrainer(
       model=model,
       train_dataset=ds["train"],
       eval_dataset=ds["test"],
       loss_class=CosineSimilarityLoss,
       metric="accuracy",
       batch_size=64,
       num_iterations=10, # The number of text pairs to generate for contrastive learning
       num_epochs=3, # The number of epochs to use for constrastive learning
       learning_rate=1.25e-05,
       column_mapping={"text": "text", "label": "label"} # Map dataset columns to text/label expected by trainer
 )

trainer.freeze()

trainer.unfreeze(keep_body_frozen=True)

trainer.train(
       num_epochs=25, # The number of epochs to train the head or the whole model (body and head)
       batch_size=64,
       body_learning_rate=1.25e-5, # The body's learning rate
       learning_rate=1e-3, # The head's learning rate
       l2_weight=0.0, # Weight decay on **both** the body and head. If `None`, will use 0.01.
   )
```

**Error:**

```python
File ~/github.com/julep-ai/monorepo/lab/.venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1190, in Module._call_impl(self, *input, **kwargs)
   1186 # If we don't have any hooks, we want to skip the rest of the logic in
   1187 # this function, and just call forward.
   1188 if not (self._backward_hooks or self._forward_hooks or self._forward_pre_hooks or _global_backward_hooks
   1189         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1190     return forward_call(*input, **kwargs)
   1191 # Do not call functions when jit is used
   1192 full_backward_hooks, non_full_backward_hooks = [], []

File ~/github.com/julep-ai/monorepo/lab/.venv/lib/python3.8/site-packages/torch/nn/modules/linear.py:114, in Linear.forward(self, input)
    113 def forward(self, input: Tensor) -> Tensor:
--> 114     return F.linear(input, self.weight, self.bias)

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument mat1 in method wrapper_addmm)
```

**What fixed it:**

```python
[...]

PR #175, opened 2022-11-12
Bugfix/body and head on different devices

blakechi:
This PR is opened for resolving #169 

- Put `model_body` on to `_target_device` right after initialization.
- Add a test to check whether the body and head are on the same device

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -164,7 +164,7 @@ class SetFitHead(models.Dense):
     def device(self) -> torch.device:
         """
         `torch.device`: The device on which the model is placed.
-        
+
         Reference from: https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L869
         """
         return next(self.parameters()).device


tests/test_modeling.py
```diff
@@ -137,6 +140,11 @@ def _build_model(num_classes: int) -> SetFitModel:
 
         return model
 
+    def test_setfit_body_and_head_on_same_device(self):
+        model = self._build_model(num_classes=1)
+
```
lewtun, line 145: Nit:
```suggestion
```

src/setfit/modeling.py
```diff
@@ -160,13 +160,20 @@ def get_loss_fn(self):
         else:
             return torch.nn.CrossEntropyLoss()
 
+    @property
+    def device(self) -> torch.device:
+        """
+        Ref from: https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L869
```
lewtun, line 166: Nit:
```suggestion
        `torch.device`: The device on which the model is placed.
        
        Reference from: https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L869
```

lewtun (approved):
Thanks for the fix @blakechi and for adding a regression test! I've left some nits, but otherwise this looks great 🚀

lewtun (approved):
Thanks for iterating @blakechi - this looks great 🔥 !

Issue #172, opened 2022-11-10
How to use max_length (number max of tokens) as an argument?


piegu:
Hi,

`max_length` is defined [here ](https://github.com/huggingface/setfit/blob/00aa8764282741b68cac4565d2f1553ecd995ae9/src/setfit/data.py#L239) with the value 32.

How to modify it?

Thank you.

blakechi:
Hello @piegu , thank you for raising up this issue.

So the `max_length` will be set to the maximum possible length according to the `model_body` (`sentence-transformers`) you use. Please see [here](https://github.com/huggingface/setfit/blob/main/src/setfit/modeling.py#L252) for details.

Hope this resolve your issue. If not, I think we can add one more argument when initializing `SetFitModel` with an assert checking whether the `max_length` exceeds the acceptable length the `model_body` can handle.

piegu:
Hi @blakechi,

Thank you for your reply.
However, I don't want to set `max_length` to the maximum possible length. I want to be able to set it to any value via an argument.

Example: texts in my dataset have a maximum of 128 tokens. Then, I want to be able to set this value as `max_length`.

Note: when using `transformers `from Hugging Face, I configure the `max_length` value in the tokenizer through the argument `max_length`.

blakechi:
Sure, I think we can achieve it as what I mentioned here:

> Hope this resolve your issue. If not, I think we can add one more argument when initializing `SetFitModel` with an assert checking whether the given `max_length` exceeds the acceptable one the `model_body` can handle.

@lewtun  How do you think? I can open a PR for this 😀

PR #176, opened 2022-11-13
Allow setting max length

blakechi:
This PR is opened for resolving #172 .

- Allow users to set `max_length` when calling `SetFitTrainer.train`.
- `max_length` will be set to the maximum number of tokens the model body can handle if it's too large.
- Add relevant tests.

A snippet of usage:
```python

num_classes = 2
train_dataset = dataset["train"].shuffle(seed=42).select(range(8 * num_classes))

    use_differentiable_head=True,
    head_params={"out_features": num_classes},
)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20,
    num_epochs=1,
    column_mapping={"sentence": "text", "label": "label"}
)

# Train the differentiable head only

# Train with custom `max_length`
trainer.train(
    num_epochs=1,
    batch_size=16,
    body_learning_rate=1e-5,
    learning_rate=1e-2,
    l2_weight=0.0,
    max_length=64,  # set your preferred length here
)
```

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -282,7 +282,7 @@ class SetFitTrainer:
                 Temporary change the weight of L2 regularization for SetFitModel's differentiable head in logistic regression.
             max_length (int, *optional*, defaults to `None`):
                 The maximum number of tokens for one data sample. Currently only for training the differentiable head.
-                If`None`, will use the maximum number of tokens the model body can accept.
+                If `None`, will use the maximum number of tokens the model body can accept.
                 If `max_length` is greater than the maximum number of acceptable tokens the model body can accept, it will be set to the maximum number of acceptable tokens.
             trial (`optuna.Trial` or `Dict[str, Any]`, *optional*):
                 The trial run or the hyperparameter dictionary for hyperparameter search.


src/setfit/modeling.py
```diff
@@ -243,13 +245,29 @@ def fit(
...
+        x_train: List[str],
+        y_train: List[int],
+        batch_size: int,
+        max_length: Optional[int] = None,
+        shuffle: bool = True,
     ) -> DataLoader:
+        max_acceptable_length = self.model_body.get_max_seq_length()
+        max_length = max_length or max_acceptable_length
+        if max_length > max_acceptable_length:
+            warnings.warn(
```
lewtun, line 258: I suggest using the logger here, e.g. 
```suggestion
            logger.warning(
```

src/setfit/modeling.py
```diff
@@ -243,13 +245,29 @@ def fit(
...
+                (
+                    f"The specified `max_length`: {max_length} is greater than the maximum length of the current model body: {max_acceptable_length}. "
+                    f"Change `max_length` to {max_acceptable_length}."
```
lewtun, line 261: Since we're overwriting the user input, should we indicate that explicitly? Alternatively, we can raise an error if you think that's a better approach
```suggestion
                    f"Using {max_acceptable_length} instead."
```

tests/test_trainer.py
```diff
@@ -172,6 +172,53 @@ def test_trainer_raises_error_with_wrong_warmup_proportion(self):
...
+    def test_trainer_max_length_exceeds_max_acceptable_length(self):
+        trainer = SetFitTrainer(
+            model=self.model,
+            train_dataset=self.dataset,
+            eval_dataset=self.dataset,
+            num_iterations=self.num_iterations,
+            column_mapping={"text_new": "text", "label_new": "label"},
+        )
+        trainer.unfreeze(keep_body_frozen=True)
+        trainer.train(
```
lewtun, line 196: Should we think about a way to actually test the large value has been overwritten with the model max length?

tests/test_trainer.py
```diff
@@ -172,6 +172,53 @@ def test_trainer_raises_error_with_wrong_warmup_proportion(self):
...
+    def test_trainer_max_length_is_smaller_than_max_acceptable_length(self):
+        trainer = SetFitTrainer(
+            model=self.model,
+            train_dataset=self.dataset,
+            eval_dataset=self.dataset,
+            num_iterations=self.num_iterations,
+            column_mapping={"text_new": "text", "label_new": "label"},
+        )
+        trainer.unfreeze(keep_body_frozen=True)
+        trainer.train(
```
lewtun, line 213: Same comment here - should we be testing the behaviour explicitly?

lewtun (commented):
Thanks for this nice quality of life improvement @blakechi 🔥 !

The PR is looking great, and I've left a few small comments that would be nice to address before we merge

tests/test_trainer.py
blakechi: you're right. I will modify the test

blakechi: just added it

src/setfit/modeling.py
```diff
@@ -243,13 +244,29 @@ def fit(
...
-        self, x_train: List[str], y_train: List[int], batch_size: int, shuffle: bool = True
+        self,
+        x_train: List[str],
+        y_train: List[int],
+        batch_size: int,
+        max_length: Optional[int] = None,
+        shuffle: bool = True,
     ) -> DataLoader:
+        max_acceptable_length = self.model_body.get_max_seq_length()
+        max_length = max_length or max_acceptable_length
```
lewtun, line 255: I think it would be nice to be explicit here:

```suggestion
        if max_length is None:
            max_length = max_acceptable_length
```

Perhaps we can also log a warning that we're using `max_acceptable_length` in this case?

tests/test_trainer.py
```diff
@@ -172,6 +177,74 @@ def test_trainer_raises_error_with_wrong_warmup_proportion(self):
...
+            max_length = 4096
+            max_acceptable_length = self.model.model_body.get_max_seq_length()
+            trainer.train(
+                num_epochs=1,
+                batch_size=3,
+                learning_rate=1e-2,
+                l2_weight=0.0,
+                max_length=max_length,
+            )
+            self.assertEqual(
```
lewtun, line 211: nice!

lewtun (approved):
Sorry for the delay and thanks for iterating @blakechi 🤘 

I've left one nit and then think we can merge!

src/setfit/modeling.py
blakechi: good point!

blakechi:
> Sorry for the delay and thanks for iterating @blakechi 🤘
> 
> I've left one nit and then think we can merge!

No worries at all! Thanks for the review 🙏🏻

src/setfit/trainer.py
```diff
@@ -279,6 +280,10 @@ def train(
                 If ignore, will be the same as `learning_rate`.
             l2_weight (float, *optional*):
                 Temporary change the weight of L2 regularization for SetFitModel's differentiable head in logistic regression.
+            max_length (int, *optional*, defaults to `None`):
+                The maximum number of tokens for one data sample. Currently only for training the differentiable head.
+                If`None`, will use the maximum number of tokens the model body can accept.
```
PhilipMay, line 285: After "If" there is a space missing.

blakechi: Thanks for finding that, @PhilipMay!

lewtun (approved):
Thanks for iterating - looks great!

Issue #135, opened 2022-10-30
Normalize before using LogisticRegression


PhilipMay:
Hi,

as far as I can see is that setfit applies LogisticRegression on top of the output of the sentence transformer model.
See here:

https://github.com/huggingface/setfit/blob/7735e8e3b208edb8dfb549beb16e585453c5f44e/src/setfit/modeling.py#L49-L55

The problem I see is that the output is not normalized by default. Since we use Cosine Sim. to compare embeddings the
length of the vector does not matter. When you do Cosine Sim. this is ok but **it is IMO not ok** when you apply LogisticRegression.

IMO the embeddings should be normalized to unit length before LogisticRegression is applied. That would be done by passing 
`normalize_embeddings=True` to the `encode` function. See here:

https://github.com/UKPLab/sentence-transformers/blob/0422a5e07a5a998948721dea435235b342a9f610/sentence_transformers/SentenceTransformer.py#L111-L118

What do you think? I can provide a PR if wanted.

doubianimehdi:
That explains those messages i'm getting : when using fit and trying to predict : 

`_logistic.py:444:` ConvergenceWarning: lbfgs failed to converge (status=1):
STOP: TOTAL NO. of ITERATIONS REACHED LIMIT.

Increase the number of iterations (max_iter) or scale the data as shown in:
    https://scikit-learn.org/stable/modules/preprocessing.html
Please also refer to the documentation for alternative solver options:
    https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression
  n_iter_i = `_check_optimize_result(`

`df_ri_manclassif['predicted']= model(df_ri_manclassif['global_text'].to_list())
this is the message : ---------------------------------------------------------------------------
NotFittedError Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_7856/2981536964.py in
----> 1 df_ri_manclassif['predicted']= model(df_ri_manclassif['global_text'].to_list())

c:\Users\doub2420\AppData\Local\Programs\Python\Python39\lib\site-packages\setfit\modeling.py in call(self, inputs)
60 def call(self, inputs):
61 embeddings = self.model_body.encode(inputs)
---> 62 return self.model_head.predict(embeddings)
63
64 def _save_pretrained(self, save_directory):

c:\Users\doub2420\AppData\Local\Programs\Python\Python39\lib\site-packages\sklearn\linear_model_base.py in predict(self, X)
445 Vector containing the class labels for each sample.
446 """
[...]

PR #177, opened 2022-11-13
add option to normalize embeddings

PhilipMay:
Add option to normalize embeddings.

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -232,6 +232,8 @@ class SetFitModel(PyTorchModelHubMixin):
                     labels = labels.to(device)
 
                     outputs = self.model_body(features)
+                    if self.normalize_embeddings:
+                        outputs = torch.nn.functional.normalize(outputs, p=2, dim=1)
                     outputs = self.model_head(outputs)
                     predictions = outputs["prediction"]
 


PhilipMay:
@blakechi do you have an idea how to best add this to the differentiable head ?

Somewhere here:

https://github.com/huggingface/setfit/blob/5313939b638b57668a1b5a67ef600e54da221d9a/src/setfit/modeling.py#L232

blakechi:
Hi @PhilipMay,

Sorry for the late reply. 
I notice that `sentence-transformers`'s `encode` function sets the model to `eval` mode at [here](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L134).
Since users might want to train `SetFitModel` in the end-to-end fashion, I think it's not a good idea to use `encode`. Instead, we might need to normalize by ourself.

It's not totally the same case but I tried adding a `LayerNorm` between the body and head before. And I found that it degraded the performance. Just for reference. 😀

PhilipMay:
@blakechi what do you think about the norm. solution I added to this PR? Can you please check it?

I just stole it form sbert. See:

https://github.com/UKPLab/sentence-transformers/blob/0b5ef4be93d2b21de3a918a084b48aab6ba48595/sentence_transformers/SentenceTransformer.py#L183-L184

blakechi (commented):
It looks good to me :)

blakechi:
Just curious, have you tested the performance? Or do you have other purposes?

PhilipMay:
> Just curious, have you tested the performance? Or do you have other purposes?

I tested performance with the "normal" Logistic Regression from SKlearn.
Normalizing reduces performance - which is surprising for me. Also see here:

https://github.com/huggingface/setfit/issues/135#issuecomment-1297000383

@blakechi do you have an idea why?

PS: To test the diff. head is still on my todo list.

PhilipMay:
Can be merged from my point of view.

blakechi:
> > Just curious, have you tested the performance? Or do you have other purposes?
> 
> I tested performance with the "normal" Logistic Regression from SKlearn. Normalizing reduces performance - which is surprising for me. Also see here:
> 
> [#135 (comment)](https://github.com/huggingface/setfit/issues/135#issuecomment-1297000383)
> 
> @blakechi do you have an idea why?
> 
> PS: To test the diff. head is still on my todo list.

I don't have a clear answer. But my guess is the magnitude of the embeddings might contains some information that helps classification, like the occurring frequency of each words?
An interesting [report](https://arxiv.org/pdf/1508.02297.pdf).


Btw, is it better to open embedding normalization to users only when we find merits from it?

PhilipMay:
> Btw, is it better to open embedding normalization to users only when we find merits from it?

I would say that ML is a field where experimentation needs to be done. You never know what exactly helps and what not.
It also depends a lot on the use-case & data. So this enables ppl. to experiment. Since we disable it by default we do no harm. We just add one more option to experiment.

lewtun (approved):
Thanks for adding this option @PhilipMay ! Since it doesn't interfere with the existing training,  I'm happy to include it :)

blakechi:
> I would say that ML is a field where experimentation needs to be done. You never know what exactly helps and what not. It also depends a lot on the use-case & data. So this enables ppl. to experiment. Since we disable it by default we do no harm. We just add one more option to experiment.

Good point, that's what I didn't notice and it makes sense to me!
Thanks for letting me know 😄 @PhilipMay

PR #183, opened 2022-11-15
delete duplicated code

Yongtae723:
Hi

I read almost all the code of setfit and found a few mistakes. so I send several PRs for each mistake. 

First, there are duplicate codes.
Since `batch_size` is already defined in two lines above, I think this is no longer necessary.


tomaarsen (approved):
LGTM

lewtun (approved):
Nice catch, thanks for the fix!

PR #184, opened 2022-11-16
Fix/input type hint

Yongtae723:
I think the type hint of predict and predict_proba is not correct.
It's python, so it doesn't cause errors, but it might confuse readers, so made this PR.

The reasons for this PR are as follows

Regarding inputs:
1. `SentenceTransformer.encode` accept `Union[List[str], str]`, not `torch.Tensor` as you can see [here](https://github.com/UKPLab/sentence-transformers/blob/0b5ef4be93d2b21de3a918a084b48aab6ba48595/sentence_transformers/SentenceTransformer.py#L111)
2.  If the type of x_test is `str`, errors occur since the size of embedding is different from the size expected by the model head.
From type these reason, the type of x_test should be   `List[str]`

As for the output, the reason is clear, so I will omit it.

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -310,7 +310,6 @@ class SetFitModel(PyTorchModelHubMixin):
         return self.model_head.predict(embeddings)
 
     def predict_proba(self, x_test: List[str]) -> Union[torch.Tensor, np.ndarray]:
-
         embeddings = self.model_body.encode(x_test, normalize_embeddings=self.normalize_embeddings)
         return self.model_head.predict_proba(embeddings)
 


tomaarsen (approved):
LGTM, well spotted

lewtun (approved):
Thanks for the fix!

lewtun:
Ah, there's some conflicts from previous PRs getting merged - would you mind resolving them and pushing again @Yongtae723 ?

Yongtae723:
Absolutely no problem

Yongtae723:
@lewtun 
Hi lewtun

I resolved it. please merge it if there are no problems.

thank you
Yongtae

lewtun:
Thanks for the fix @Yongtae723!

PR #187, opened 2022-11-16
Resolve different output shape of differentiable head

### LINKED ISSUE #186: [Bug] `__call__`/`predict` has different shape using differentiable head
Hello!

## Bug overview
* When using a differentiable head, the output shape of calling `__call__` or `predict` (they're aliases) is `(N, 1)` with `N` as the number of parameters.
* When using the logistic regression head, the output of that same call is `(N,)`.

### For a differentiable head

<details><summary>Copy-pasteable reproducing script (Script 1)</summary>

```python

    use_differentiable_head=True,
)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20,  # The number of text pairs to generate for contrastive learning
    num_epochs=1,  # The number of epochs to use for constrastive learning
    column_mapping={"sentence": "text", "label": "label"}  # Map dataset columns to text/label expected by trainer
)

# Train and evaluate
trainer.train()

# Run inference
preds = model(["i loved the spiderman movie!", "pineapple on pizza is the worst 🤮"])
print(preds.shape)
```
</details>

```
(2, 1)
```
Note that we don't need to train this model to show the bug. 

### For a logistic regression head
<details><summary>Copy-pasteable reproducing script (Script 2)</summary>

```python

)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20,  # The number of text pairs to generate for contrastive learning
    num_epochs=1,  # The number of epochs to use for constrastive learning
    column_mapping={"sentence": "text", "label": "label"}  # Map dataset columns to text/label expected by trainer
)

# Train and evaluate
trainer.train()

# Run inference
preds = model(["i loved the spiderman movie!", "pineapple on pizza is the worst 🤮"])
print(preds.shape)
```
</details>

```
(2,)
```
Note that the `LogisticRegression` head must be fitted before we can call `__call__`, hence the larger example. This example is nearly directly taken from the `README.md`.

## Expected behaviour
I would expect the output shape to be `(2,)` in both situations.

## Consequences
This is a fairly easy fix, but I found it curious that this was not discovered earlier. After all, the predictions from `SetFitHead` should be used in a loss function when training, right? An issue in the shape would be discovered there. Through some testing, I found out that the following fitting loop for the differentiable head is **never called** with the script above for the differentiable head:
https://github.com/huggingface/setfit/blob/4a613b08267690dad0840444fc3e2caf60f29a44/src/setfit/modeling.py#L214-L240

If instead I use the Trainer like recommended from the following training script, with `keep_body_frozen=True`, then I do get an error.

https://github.com/huggingface/setfit/blob/4a613b08267690dad0840444fc3e2caf60f29a44/scripts/setfit/run_fewshot.py#L132-L142

<details><summary>Copy-pasteable reproducing script (Script 3)</summary>

```python

    use_differentiable_head=True
)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20,  # The number of text pairs to generate for contrastive learning
    num_epochs=1,  # The number of epochs to use for constrastive learning
    column_mapping={"sentence": "text", "label": "label"}  # Map dataset columns to text/label expected by trainer
)

# Freeze head
trainer.freeze()

# Do contrastive training
trainer.train()

# Unfreeze head and freeze body

# Train end-to-end
trainer.train(
    num_epochs=25,
    body_learning_rate=1e-5,
    learning_rate=1e-2,
    l2_weight=0.0,
)

# Run inference
preds = model(["i loved the spiderman movie!", "pineapple on pizza is the worst 🤮"])
print(preds)
```
</details>

The error is the following:
```python
  File "[sic]\setfit\demo_diff_head_freeze.py", line 42, in <module>
    trainer.train(
  File "[sic]\setfit\src\setfit\trainer.py", line 376, in train
    self.model.fit(
  File "[sic]\setfit\src\setfit\modeling.py", line 236, in fit
    loss = criterion(predictions, labels)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\modules\module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\modules\loss.py", line 619, in forward
    return F.binary_cross_entropy(input, target, weight=self.weight, reduction=self.reduction)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\functional.py", line 3086, in binary_cross_entropy
    raise ValueError(
ValueError: Using a target size (torch.Size([16])) that is different to the input size (torch.Size([16, 1])) is deprecated. Please ensure they have the same size.
```

Triggering training on the differentiable head is very confusing right now, and so I would like to continue the discussion at https://github.com/huggingface/setfit/issues/179#issuecomment-1316710169 on a new class structure.

I'll submit a PR for this shortly.

- Tom Aarsen

tomaarsen:
Resolves #186

Hello!

## Pull request overview
* Fix issue causing the output shape of `__call__` of the differentiable head to be `(N, 1)` rather than the expected `(N,)`
* Fix type mismatch issue preventing fitting the differentiable head.

## Details
This PR contains two separate fixes. 

### Fix one
The first one is the inclusion of `.squeeze()` in the `SetFitHead` forward call, on these lines:
https://github.com/huggingface/setfit/blob/4a613b08267690dad0840444fc3e2caf60f29a44/src/setfit/modeling.py#L121-L122

`x` has shape `[num_labels, embedding_length]`, e.g. `[2, 768]`, causing the `logits` to have shape `[num_labels, 1]`, e.g. `[2, 1]`.
This shape is then continued and eventually returned, causing the first bug reported in #186.
The `.squeeze()` call squeezes the logits shape from `[num_labels, 1]` to `[num_labels]`.

### Consequences of fix one
After this fix, Script 1 from #186 works like intended, resulting in an output shape of `(2,)`.

Furthermore, the error from Script 3 from #186 has changed. Rather than complaining about a shape mismatch, it now complains about a type mismatch:
```python
  File "[sic]\setfit\demo_diff_head_freeze.py", line 42, in <module>
    trainer.train()
  File "[sic]\setfit\src\setfit\trainer.py", line 376, in train
    self.model.fit(
  File "[sic]\setfit\src\setfit\modeling.py", line 236, in fit
    loss = criterion(predictions, labels)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\modules\module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\modules\loss.py", line 619, in forward
    return F.binary_cross_entropy(input, target, weight=self.weight, reduction=self.reduction)
  File "[sic]\envs\setfit\lib\site-packages\torch\nn\functional.py", line 3095, in binary_cross_entropy
    return torch._C._nn.binary_cross_entropy(input, target, weight, reduction_enum)
RuntimeError: Found dtype Long but expected Float
```

### Fix two
The second fix addresses the error for Script 3 after fix 1. The predictions and labels passed to the loss function are now the same shape, but the predictions are float, while the labels are longs. The latter is because of the following line:
https://github.com/huggingface/setfit/blob/4a613b08267690dad0840444fc3e2caf60f29a44/src/setfit/data.py#L280

I've changed `.long()` to `.float()`. 

### Consequences of fix two
After this fix, both of the errors reported from #186 are resolved, and the outputs are exactly like expected. Using Script 3 from #186, it is now possible to train the head.

---

cc: @lewtun 
cc: @blakechi for your familiarity with the differentiable head

- Tom Aarsen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ pypi_test_upload:
 	twine upload dist/* -r pypitest --repository-url=https://test.pypi.org/legacy/
 
 pypi_test_install:
-	python -m pip install evaluate==0.2.2 datasets==2.3.2 sentence_transformers==2.2.2
+	python -m pip install evaluate==0.3.0 datasets==2.3.2 sentence_transformers==2.2.2
 	python -m pip install -i https://testpypi.python.org/pypi setfit
 	python -c "from setfit import *"
 	echo "🚀 Successfully installed setfit from test.pypi.org"
--- a/scripts/tfew/requirements.txt
+++ b/scripts/tfew/requirements.txt
@@ -9,5 +9,5 @@ sentencepiece==0.1.96
 scipy
 ipdb
 evaluate==0.1.2
-sklearn
+scikit-learn
 git+https://github.com/SetFit/promptsource.git
--- a/scripts/transformers/requirements.txt
+++ b/scripts/transformers/requirements.txt
@@ -1,4 +1,4 @@
 transformers[sentencepiece,optuna]==4.20.0
 torch==1.11
-sklearn
+scikit-learn
 typer
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ README_TEXT = (Path(__file__).parent / "README.md").read_text(encoding="utf-8")
 
 INTEGRATIONS_REQUIRE = ["optuna"]
 
-REQUIRED_PKGS = ["datasets==2.3.2", "sentence-transformers==2.2.2", "evaluate==0.2.2"]
+REQUIRED_PKGS = ["datasets==2.3.2", "sentence-transformers==2.2.2", "evaluate==0.3.0"]
 
 QUALITY_REQUIRE = ["black", "flake8", "isort", "tabulate"]
 
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -89,9 +89,9 @@ class SetFitHead(models.Dense):
         self.out_features = out_features
         self.temperature = temperature
         self.bias = bias
-        self.device = device or "cuda" if torch.cuda.is_available() else "cpu"
+        self._device = device or "cuda" if torch.cuda.is_available() else "cpu"
 
-        self.to(device)
+        self.to(self._device)
         self.apply(self._init_weight)
 
     def forward(
@@ -160,13 +160,22 @@ class SetFitHead(models.Dense):
         else:
             return torch.nn.CrossEntropyLoss()
 
+    @property
+    def device(self) -> torch.device:
+        """
+        `torch.device`: The device on which the model is placed.
+
+        Reference from: https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L869
+        """
+        return next(self.parameters()).device
+
     def get_config_dict(self) -> Dict[str, Union[int, float, bool]]:
         return {
             "in_features": self.in_features,
             "out_features": self.out_features,
             "temperature": self.temperature,
             "bias": self.bias,
-            "device": self.device,
+            "device": self.device.type,  # store the string of the device, instead of `torch.device`
         }
 
     @staticmethod
@@ -190,6 +199,7 @@ class SetFitModel(PyTorchModelHubMixin):
         model_head: Optional[Union[nn.Module, LogisticRegression]] = None,
         multi_target_strategy: str = None,
         l2_weight: float = 1e-2,
+        normalize_embeddings: bool = False,
     ) -> None:
         super(SetFitModel, self).__init__()
         self.model_body = model_body
@@ -199,6 +209,7 @@ class SetFitModel(PyTorchModelHubMixin):
         self.l2_weight = l2_weight
 
         self.model_original_state = copy.deepcopy(self.model_body.state_dict())
+        self.normalize_embeddings = normalize_embeddings
 
     def fit(
         self,
@@ -230,6 +241,8 @@ class SetFitModel(PyTorchModelHubMixin):
                     labels = labels.to(device)
 
                     outputs = self.model_body(features)
+                    if self.normalize_embeddings:
+                        outputs = torch.nn.functional.normalize(outputs, p=2, dim=1)
                     outputs = self.model_head(outputs)
                     predictions = outputs["prediction"]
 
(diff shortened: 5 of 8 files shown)


tomaarsen:
As it turns out, fix 2 is a bit tricky due to the different requirements of the different loss functions. The input (i.e. the prediction) is always `float32`, so that's all sorted. Then, the `BCELoss` wants the target (i.e. the labels) to be the same dtype as the input, so also `float32`. This is why I've opted to make fix 2.

However, the other loss used by the `SetFitHead`, `CrossEntropyLoss`, wants the target to be `float64` whenever the target has a different shape than the input.

It seems that always casting to either `float32` or `float64` will cause issues with one of the two losses.

@blakechi Do you have a recommendation here? Obviously, we can simply cast like `labels.float() if isinstance(criterion, nn.BCELoss) else labels.long()`, but surely there has to be a more elegant solution.

blakechi:
Hi @tomaarsen,

My opinion is how about we use only `CrossEntropyLoss` for binary and multi-class classification? So we don't need to worry about the type. Check out [here](https://github.com/huggingface/setfit/issues/186#issuecomment-1319557707) in #187. Thanks!

PhilipMay:
> My opinion is how about we use only `CrossEntropyLoss` for binary and multi-class classification? So we don't need to worry about the type. Check out [here](https://github.com/huggingface/setfit/issues/186#issuecomment-1319557707) in #187. Thanks!

AFAIK that is also how Hugging Face Transformers is doing their classification. They do not have special cases for binary classes and multi classes.

blakechi:
> > My opinion is how about we use only `CrossEntropyLoss` for binary and multi-class classification? So we don't need to worry about the type. Check out [here](https://github.com/huggingface/setfit/issues/186#issuecomment-1319557707) in #187. Thanks!
> 
> AFAIK that is also how Hugging Face Transformers is doing their classification. They do not have special cases for binary classes and multi classes.

Thanks for pointing that out!

tomaarsen:
On hold until #203 is merged!

tomaarsen:
Superseded by #203

PR #194, opened 2022-11-17
Update evaluate

lvwerra:
There is an issue with the hub integration before 0.3.0 and evaluate which might cause this version to break in a few months.


lewtun (approved):
Thanks for the fix!

PR #196, opened 2022-11-18
add option to set `samples_per_label`

PhilipMay:
This PR adds the option to set `samples_per_label` in the `SentenceLabelDataset` when you train with triplet loss.

See here in sbert doc:
https://github.com/UKPLab/sentence-transformers/blob/0b5ef4be93d2b21de3a918a084b48aab6ba48595/sentence_transformers/datasets/SentenceLabelDataset.py#L31-L32


lewtun (approved):
Thanks for adding this option @PhilipMay - LGTM!

PR #200, opened 2022-11-22
Use scikit-learn rather than sklearn in requirements files

lesteve:
scikit-learn is the actual package name, using `sklearn` in requirement files will stop working December 1st 2022. For more details, see https://github.com/scikit-learn/sklearn-pypi-package.


lewtun (approved):
Good catch, thanks for the fix!

PR #201, opened 2022-11-23
No longer needlessly deepcopy the original model state

### LINKED ISSUE #199: Question : why is model_original_state  necessary?
In SetFitModel, self.model_original_state is defined, as you can see [here](https://github.com/huggingface/setfit/blob/main/src/setfit/modeling.py#L225:~:text=l2_weight%20%3D%20l2_weight-,self.model_original_state%20%3D%20copy.deepcopy(self.model_body.state_dict()),-def%20fit).

But I don't see this variable being used again. So I wonder why you define that.
Since self.model_original_state contains the weights of sentence transformers, which leads to a heavy model class, so if it is not necessary, we can delete it to make a light model.

Could you tell me your opinion?

Thank you.
I am looking forward to your reply.

Yongtae

tomaarsen:
Resolves #199 

Hello!

## Pull request overview
* Remove `deepcopy` of `model.state_dict`.
* Modify [`scripts/setfit/run_fewshot_multilabel.py`](https://github.com/huggingface/setfit/blob/main/scripts/setfit/run_fewshot_multilabel.py) to reflect that `model.model_original_state` has been removed.

## Details
As mentioned by @Yongtae723 in #199, the model state is copied and then [never used again](https://github.com/huggingface/setfit/search?q=model_original_state), with exception of `scripts/setfit/run_fewshot_multilabel.py`. This is an unnecessary waste of memory, and hence we believe that these lines should be removed.

Furthermore, I modified `scripts/setfit/run_fewshot_multilabel.py` to perform the deepcopying in that script itself. Note that this is somewhat different than the original situation, as now these two lines are executed *before* the deepcopying is performed, rather than the deepcopy occurring first:
https://github.com/huggingface/setfit/blob/b66760cadb5afa27b9dc5b6ee1406b170be06568/src/setfit/modeling.py#L47-L48

I suspect that this won't matter too much, but I'll leave that up to your judgement.
The script still works like before, e.g. using:
```shell
python scripts/setfit/run_fewshot_multilabel.py --sample_sizes 64 --datasets=go_emotions
```

- Tom Aarsen

### DIFF
--- a/scripts/setfit/run_fewshot_multilabel.py
+++ b/scripts/setfit/run_fewshot_multilabel.py
@@ -100,6 +100,7 @@ class RunFewShot:
             self.args.model, max_seq_length=args.max_seq_length, add_normalization_layer=args.add_normalization_layer
         )
         self.model = self.model_wrapper.model
+        self.model_original_state = copy.deepcopy(self.model.state_dict())
 
     def get_classifier(self, sbert_model: SentenceTransformer) -> SKLearnWrapper:
         if self.args.classifier == "logistic_regression":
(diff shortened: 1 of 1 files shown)


Yongtae723:
@tomaarsen 
Hi Tom!

thank you for making excellent modifications and PR!

tomaarsen:
And thanks to you for spotting this issue!

blakechi (approved):
Looks good to me. Nice work!

Issue #179, opened 2022-11-13
Refactor training hyperparameters in `SetFitTrainer` to dedicated class


lewtun:
The current implementation of `SetFitTrainer` has become a bit bloated because it accepts all the hyperparameters for training as separate arguments. In hindsight, it would have been better to collect all the hyperparameters in a separate data class (similar to the [`TrainingArguments`](https://huggingface.co/docs/transformers/v4.24.0/en/main_classes/trainer#transformers.TrainingArguments) of `transformers`). At the same time, it doesn't seem necessary to prepend `SetFit` to the name of these classes.

To that end, it would be great to:

- [ ] Create new `TrainingArguments` and `Trainer` classes, where `Trainer` accepts all the hyperparameters as an `args` parameter.
- [ ] Add a deprecation notice on `SetFitTrainer` to inform users about the new `Trainer` (and let them know it will be removed in 2 minor releases time)

tomaarsen:
I'll get working on this in a week on the 21st, if it hasn't been claimed by then!

- Tom Aarsen

lewtun:
Thanks @tomaarsen - that would be wonderful! Feel free to ping me if you need any guidance :)

tomaarsen:
Now that `DistillationSetFitTrainer` has entered the picture, how do we want to structure the new classes?
`Trainer`, `DistillationTrainer` and `TrainingArguments`?

Furthermore, I noticed that training using `use_differentiable_head=True` requires some freezing/unfreezing and multiple calls to `trainer.train()`. My intuition for a `Trainer` is that it requires one call to perform all of the training, which is not what we're seeing now. I also noticed that the differentiable head training does not allow flexibility in all parameters, like the optimizer, the scheduler, and its parameters. 

If we indeed want to allow one call to train the model fully, then we may want to create the following structure:

```python
@dataclass
class TrainingArguments:
    body_learning_rate = 2e-5
    num_iterations = 20
    num_epochs = 1
    batch_size = 16
    l2_weight = 0.0
    grad_accum_steps = 4
    loss_class = losses.CosineSimilarityLoss
    scheduler: str = "WarmupLinear"
    optimizer_class = AdamW
    ...


class BodyTrainingArguments(TrainingArguments):
    pass


class HeadTrainingArguments(TrainingArguments):
    loss_class = losses.CosineSimilarityLoss
    body_learning_rate = 1e-5
    head_learning_rate = 1e-2
    num_epochs = 25
    freeze_body = False
[...]

PR #203, opened 2022-11-24
Feature/deprecate binary cross entropy loss

blakechi:
This PR is opened to resolve an [issue](https://github.com/huggingface/setfit/pull/187#issuecomment-1317112370) found by @tomaarsen. Thanks @tomaarsen and nice PR #187, I use it as the template!

### Major changes
- Deprecate `torch.nn.BCELoss` for binary classification and replace it by `torch.nn.CrossEntropyLoss` as 2-class classification.
- Update docstring
- Update `test_modeling.py`

### Minor changes
- Add `eps` for numerical stability when scaling `logits` with `temperature`

### Problem
As @tomaarsen mentioned [here](https://github.com/huggingface/setfit/pull/187#issuecomment-1317112370):
> As it turns out, fix 2 is a bit tricky due to the different requirements of the different loss functions. The input (i.e. the prediction) is always `float32`, so that's all sorted. Then, the `BCELoss` wants the target (i.e. the labels) to be the same dtype as the input, so also `float32`. This is why I've opted to make fix 2.
> 
> However, the other loss used by the `SetFitHead`, `CrossEntropyLoss`, wants the target to be `float64` whenever the target has a different shape than the input.

`BCELoss` and `CrossEntropyLoss` require different data types respectively, which increases complexity to cast the labels into correct data type and makes the code harder to read.

### Solution
Therefore, this PR deprecates `BCELoss` and use `CrossEntropyLoss` for binary classification.

### Result
I run some experiments on `CR` and `EnronSpam` datasets to check whether the performance is changed. 
Here is the results with **N=8** as **Table 2** in the paper:

| Head | CR | EnronSpam | Amazon-CF|
-- | -- | -- | --
| sklearn (from paper) | 88.5 (1.9) | 90.1 (3.4) | 40.3 (11.8)
| pytorch (BCE) | 88.8 (1.3) | 90.0 (3.3) | 76.4 (11.2)
| pytorch (CrossEntropy) | 88.8 (1.4) | 90.3 (3.2) | 73.1 (10.5)

(Hyperparamters: batch size = 16, L2 weight (weight decay) = 0, head learning rate = 1e-2, keep body frozen)

From the results, their performance are similar except **Amazon-CF**. It looked strange to me and after several runs, I still got similar results at the end. [Here](https://colab.research.google.com/drive/1pRIudDMWK4ZiLEpPkZJ3EVcbv1uF_2EI?usp=share_link) is the notebook I used to run experiments for `pytorch (CrossEntropy)`.

## Conclusion
The advantages of this change (copied from [here](https://github.com/huggingface/setfit/issues/186#issuecomment-1319557707)):

- Solve the [issue](https://github.com/huggingface/setfit/pull/187#issuecomment-1317112370) about different data types for different loss functions.
- Won't forget to test either in the future when other features add in. 
- Make #179 easier.

Still need to validate whether the results of `Amazon-CF` is correct. @lewtun could you have a check on the notebook to validate the experiments? Thanks!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,36 @@
+# Configuration file to control (pytest) coverage
+[run]
+# Run branch coverage, too
+branch = True
+
+[paths]
+source =
+    src/setfit
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code:
+    def __repr__
+    if self\.debug
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise AssertionError
+    raise NotImplementedError
+
+    # Don't complain if non-runnable code isn't run:
+    if 0:
+    if __name__ == .__main__.:
+
+    # Don't complain about abstract methods, they aren't run:
+    @(abc\.)?abstractmethod
+
+    # Ignore TYPE_CHECKING code
+    if TYPE_CHECKING:
+
+[html]
+directory = coverage_report_html
+title = SetFit coverage report
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,17 +13,37 @@ jobs:
 
   test_sampling:
     name: Run unit tests
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        os: [ubuntu-latest, windows-latest]
+        requirements: ['.[tests]', '.[compat_tests]']
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
       - name: Setup Python environment
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.7
-      - name: Install dependencies
+          python-version: ${{ matrix.python-version }}
+
+      - name: Try to load cached dependencies
+        uses: actions/cache@v3
+        id: restore-cache
+        with:
+          path: ${{ env.pythonLocation }}
+          key: python-dependencies-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.requirements }}-${{ hashFiles('setup.py') }}-${{ env.pythonLocation }}
+
+      - name: Install external dependencies on cache miss
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install ".[tests]"
+          python -m pip install --no-cache-dir --upgrade pip
+          python -m pip install --no-cache-dir ${{ matrix.requirements }}
+        if: steps.restore-cache.outputs.cache-hit != 'true'
+
+      - name: Install the checked-out setfit
+        run: python -m pip install .
+
       - name: Run unit tests
         run: pytest -sv tests/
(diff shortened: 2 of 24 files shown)


tomaarsen (approved):
This is looking great! Thanks for this. Well written PR, too!

I reran your experiments, and can corroborate the results. It all looks above board, too. Very interesting to see such large differences compared to the LogisticRegression head. Perhaps this is caused by a class inbalance or something similar?
I did some further tests with other datasets, i.e. non-binary, and it still works like intended. 

I'll put #187 on hold until this is merged.

blakechi:
Thanks for you feedbacks, @tomaarsen! That really helps validate the experiments!

I checked the test dataset of **Amazon-CF** and it matched your guess. The proportion is 10% (503 samples) for 1 vs 90% (5497 samples) for 0. Now I'm thinking why in this [PR](https://github.com/huggingface/setfit/pull/112#issuecomment-1295773537) I didn't get similar results as what we see now. Maybe something is changed? Will look into it.

tomaarsen:
@blakechi I had gotten a hunch about the results and the class imbalance. No reputable researchers would evaluate a classification task with a 90:10 class inbalance using accuracy, and most certainly no reasonable data scientist would then only score a ~43% on the task, hah!
The discrepancy is shown here:
https://github.com/huggingface/setfit/blob/fa1021d2355f0cb3a2c85732ee7ffe44b0cef0d1/src/setfit/utils.py#L27-L34

In the [paper](https://arxiv.org/pdf/2209.11055.pdf), `amazon_counterfactual_en` is evaluated using the matthews correlation coefficient, rather than accuracy. See Table 6, appendix A for the details. 
However, in the script, if you supply just a list of datasets, each will be evaluated using just accuracy. Your earlier test used the test dataset from the previous snippet, causing the AmazonCF task to be evaluated using MCC.

---

I've re-ran the experiments for the `LogisticRegression` head and the differentiable `SetFitHead` with these commands:
```
python scripts/setfit/run_fewshot.py --sample_sizes=8 --lr=0.01 --keep_body_frozen --datasets amazon_counterfactual_en --batch_size 4
```

```
python scripts/setfit/run_fewshot.py --sample_sizes=8 --classifier=pytorch --lr=0.01 --keep_body_frozen --datasets amazon_counterfactual_en --batch_size 4
```
with the following change to `scripts/setfit/run_fewshot.py`:
```diff
diff --git a/scripts/setfit/run_fewshot.py b/scripts/setfit/run_fewshot.py
index 088e25d..ae03992 100644
--- a/scripts/setfit/run_fewshot.py
+++ b/scripts/setfit/run_fewshot.py
@@ -91,7 +91,7 @@ def main():
     elif args.is_test_set:
         dataset_to_metric = TEST_DATASET_TO_METRIC
     else:
-        dataset_to_metric = {dataset: "accuracy" for dataset in args.datasets}
+        dataset_to_metric = {dataset: "matthews_correlation" for dataset in args.datasets}
 
     # Configure loss function
     loss_class = LOSS_NAME_TO_CLASS[args.loss]
```

---

This has resulted in these outputs:

| Head  | Amazon CF (MCC) |
| ------------- | ------------- |
| `LogisticRegression` | 44.4 (8.6) |
| `SetFitHead` w. `CrossEntropyLoss`  | 44.6 (16.9) |

This is much more along the line of what we would expect. With other words, this PR seems to work like intended, without any fun additional surprises.

- Tom Aarsen

blakechi:
Great finding! Ya, that explains why the numbers are so different. Appreciated for the experiments as well!

blakechi:
Put this PR on hold since #207 might make some changes that overlaps.

tomaarsen (approved):
I think this ought to be ready for merging now. I resolved some merge conflicts and verified that training after this PR works equivalently to the main branch. I ran several experiments using sst2 and sst5 with success, and clearly the tests pass, too.

Additionally, this PR should supersede #187 and solve #186.

PR #212, opened 2022-11-29
Add multi-target support to SetFitHead

### LINKED ISSUE #211: SetFitHead incompatible with multi-label classification
Hello!

I want to use `SetFitHead` in a multi-label setting but there seems to be some problems. It works well with the logistic regression head, but not with the differentiable head.

The following code yields an error
```python
dataset = load_dataset()  # "text" and "label" keys. Labels are "multiple-hot" encoded floats, from `sklearn.preprocessing.MultiLabelBinarizer`

st_model = SentenceTransformer(modules=[word_embedding_model, pooling_model])
head = SetFitHead(
    st_model.get_sentence_embedding_dimension(),
    len(dataset["train"][0]["label"]),
)
setfitmodel = SetFitModel(st_model, head, multi_target_strategy="one-vs-all")

trainer = SetFitTrainer(
    model=setfitmodel,
    train_dataset=dataset["train"],
    eval_dataset=dataset["test"],
    loss_class=losses.CosineSimilarityLoss,
    metric=compute_metrics,
)
trainer.freeze()
trainer.train()

trainer.unfreeze()
trainer.train(
    num_epochs=25,
    batch_size=16,
    body_learning_rate=1e-5,
    learning_rate=1e-2,
    l2_weight=0.0,
)
```

When running this code, the dataloader in `setfit.train` returns long labels rather than float(they get converted by the `Dataloader` even though `SetFitDataset.y` is the correct type), leading to an exception in `torch.nn.CrossEntropyLoss`. The issue could be fixed by converting the values to floats before calculating loss.

However, even after that fix, classification is not made in a multi-label fashion for the following reasons:
- Softmax activation function; should be element-wise sigmoid for multilabel classification
- Argmax of output in `SetFitHead.predict` if `out_features > 1`

I'd like a heads-up if I am doing something wrong, otherwise I could submit a PR to make `SetFitHead` work with multi-label classification.

OskarLiew:
Closes #211

- Now uses element-wise sigmoid instead of softmax if `multitarget=True` 
- Only uses argmax for single-target multi-class.
- Always allow temperature and update docstring
  - I felt like there was a typo in the docstring, since it was used for multi-class, not multi-target. However, I see no reason to limit it to softmax since the sigmoid has the same concept of temperature.
- Allow multitarget initialization with `from_pretrained` if the strategy is `"one-vs-all"`  or `"multi-output"`, since they are the same in this context (pretty sure at least) and that is what I implemented

### DIFF
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -229,7 +229,6 @@ class SetFitTrainerMultilabelDifferentiableTest(TestCase):
         multilabel_accuracy_metric = evaluate.load("accuracy", "multilabel")
 
         def compute_metrics(y_pred, y_test):
-            print(y_pred, y_test)
             return {
                 "f1": multilabel_f1_metric.compute(predictions=y_pred, references=y_test, average="micro")["f1"],
                 "accuracy": multilabel_accuracy_metric.compute(predictions=y_pred, references=y_test)["accuracy"],


src/setfit/modeling.py
```diff
@@ -76,6 +77,7 @@ def __init__(
         temperature: float = 1.0,
         bias: bool = True,
         device: Optional[Union[torch.device, str]] = None,
+        multitarget: bool = False,
```
tomaarsen, line 80: A quick nitpick: I'm missing a docstring for this.

src/setfit/modeling.py
```diff
@@ -120,10 +124,11 @@ def forward(
 
         x = features["sentence_embedding"] if is_features_dict else features
         logits = self.linear(x)
-        if self.out_features == 1:  # only has one target
-            outputs = torch.sigmoid(logits)
-        else:  # multiple targets
-            temperature = temperature or self.temperature
+
+        temperature = temperature or self.temperature
+        if self.multitarget or self.out_features == 1:  # only has one target or multiple targets per item
```
blakechi, line 129: Should also update line:163 to use `BCELoss` when using multi-label classification

blakechi (commented):
Looks good!

Some extra suggestions:
- Should also update [`SetFitDataset`](https://github.com/huggingface/setfit/blob/main/src/setfit/data.py#L217) for multi-target classification.
- It would be great if you can add some tests to test your updates [here](https://github.com/huggingface/setfit/blob/main/tests/test_modeling.py#L98)!
- (Not urgent, optional) Make `SetFitHead` be available in [run_fewshot_multilabel.py](https://github.com/huggingface/setfit/blob/main/scripts/setfit/run_fewshot_multilabel.py)

src/setfit/data.py
```diff
@@ -222,7 +225,7 @@ class SetFitDataset(TorchDataset):
     Args:
         x (`List[str]`):
             A list of input data as texts that will be fed into `SetFitModel`.
-        y (`List[int]`):
+        y (`Union[List[int], List[List[int]]]`):
```
OskarLiew, line 228: @blakechi I didn't really understand which changes to the dataset you were thinking about because it works as is, but I updated the type hints to make it more obvious that it works with multitarget labels.

src/setfit/modeling.py
```diff
@@ -132,30 +140,28 @@ def forward(
 
         return outputs
 
-    def predict_proba(self, x_test: torch.Tensor) -> torch.Tensor:
+    def predict_proba(self, x_test: Union[torch.Tensor, "ndarray"]) -> Union[torch.Tensor, "ndarray"]:
```
OskarLiew, line 143: I moved some functionality from `predict` to here to make them work the same way. I did this because I found it cumbersome to write the tests without it.

src/setfit/modeling.py
```diff
@@ -132,30 +140,28 @@ def forward(
...
-            out = torch.where(probs >= 0.5, 1, 0)
+        if self.out_features == 1 or self.multitarget:
+            out = np.where(probs >= 0.5, 1, 0)
         else:
-            out = torch.argmax(probs, dim=-1)
-
-        if not is_tensor:
-            return out.cpu().numpy()
-
+            out = np.argmax(probs, dim=-1)
```
OskarLiew, line 160: `np.where` and `np.argmax` works with both ndarrays and tensors and I double checked that they return a tensor if given a tensor, making them more versatile than the `torch` variants

src/setfit/modeling.py
```diff
@@ -239,6 +245,8 @@ def fit(
                     # to model's device
                     features = {k: v.to(device) for k, v in features.items()}
                     labels = labels.to(device)
+                    if self.model_head.multitarget:
+                        labels = labels.float()
```
OskarLiew, line 249: Added this while waiting for #187. Required for the loss function using multitarget labels

src/setfit/modeling.py
```diff
@@ -305,11 +313,11 @@ def _freeze_or_not(self, model: torch.nn.Module, to_freeze: bool) -> None:
         for param in model.parameters():
             param.requires_grad = not to_freeze
 
-    def predict(self, x_test: torch.Tensor) -> torch.Tensor:
+    def predict(self, x_test: Union[str, List[str]]) -> Union[torch.Tensor, "ndarray"]:
         embeddings = self.model_body.encode(x_test, normalize_embeddings=self.normalize_embeddings)
         return self.model_head.predict(embeddings)
 
-    def predict_proba(self, x_test: torch.Tensor) -> torch.Tensor:
+    def predict_proba(self, x_test: Union[str, List[str]]) -> Union[torch.Tensor, "ndarray"]:
```
OskarLiew, line 320: These two function signatures were misleading, since `SentenceTransformer.encode` takes a `Union[str, List[str]]`, so I updated it to be the same here.

OskarLiew (commented):
Thank you @blakechi @tomaarsen for your comments ⭐. I have updated according to your feedback. I was a bit unsure of which changes to the dataset Blake was thinking about, but I at least updated the type hints there. Please see my comments to see my thoughts on the changes and let me know if I should change something else.

src/setfit/modeling.py
tomaarsen: 👍 this change was discussed in https://github.com/huggingface/setfit/pull/207#issuecomment-1330344680 as well.

tests/test_trainer.py
```diff
@@ -212,6 +212,54 @@ def compute_metrics(y_pred, y_test):
...
+            metric=compute_metrics,
+            num_iterations=self.num_iterations,
+            column_mapping={"text_new": "text", "label_new": "label"},
+        )
+
+        trainer.freeze()
+        trainer.train()
+
+        trainer.unfreeze(keep_body_frozen=False)
+        trainer.train(5)
```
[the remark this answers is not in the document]
tomaarsen: I think it's common to have a much higher number of epochs for training the differentiable head. For example, the fewshot testing script uses 25:
https://github.com/huggingface/setfit/blob/main/scripts/setfit/run_fewshot.py#L137

I ran the test a few times and it doesn't seem to cause issues :) 👍

src/setfit/data.py
blakechi: Oh sorry, I should be more specific. I was thinking the converting for labels, which I saw you added this at line 248 and 249.
Look great now :)

src/setfit/modeling.py
blakechi: I will suggest this PR be merged before #203 and #187.
Please see the reasons [here](https://github.com/huggingface/setfit/issues/211#issuecomment-1331656303) in #211.

Yongtae723:
Hi everyone!

As I talked #267 I will fix conflict to be more great repo!

Thank you!

Yongtae723:
Hi everyone!

Has anyone ever tried to train with model_head of nn.Module?

when I do this, the accuracy is not high like sklearn.

this [example](https://colab.research.google.com/github/huggingface/setfit/blob/main/notebooks/text-classification_multilabel.ipynb) shows about accuracy of 0.4, but nn.Module generate 0.1

I want to know this is true,  or I have some error

Yongtae723:
I created New PR #271!

tomaarsen:
> Hi everyone!
> 
> Has anyone ever tried to train with model_head of nn.Module?
> 
> when I do this, the accuracy is not high like sklearn.
> 
> this [example](https://colab.research.google.com/github/huggingface/setfit/blob/main/notebooks/text-classification_multilabel.ipynb) shows about accuracy of 0.4, but nn.Module generate 0.1
> 
> I want to know this is true, or I have some error

I'm afraid you're training incorrectly. 
Please see https://github.com/huggingface/setfit/issues/271#issuecomment-1382831316

tomaarsen:
Superseded and implemented by #272. Thank you for providing the foundation for that PR, @OskarLiew!

PR #223, opened 2022-12-07
Add SetFitModelConfig

### LINKED ISSUE #124: id2label and label2id setup for inference
I have add the id2label and labnel2id top the pretrained model, I know this is a Sentence Transformer model but I dont know if there is a way to pass the label mapper to the head of the SetFitModel 

`model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2",num_labels=num_classes,id2label=id2label,label2id=label2id)
`

In that way in inference time we just call the SetFit Model

jegork:
Fixes #124

I have added a `SetFitModelConfig` and setter/getter functions for `SetFitModel` to be able to change the configuration of the underlying transformer body model.

This gives ability to provide e.g. id2label via
```python
model = SetFitModel.from_pretrained(model_name)
model.config.model_body.idx2label = {0: "neg", 1: "pos"}
```

One disadvantage of this approach is that the provided code has to access protected members of `model_body`, but I don't think there is any other way to access the `PretrainedConfig` for the `model_body`.

Additionally, in the config setter, I have currently left the possibility to change `model_body` config only (because there is currently no single config for model head). In getter I return `SetFitHead.get_config_dict` as the `model_head` config.

Maybe we can come up with a way to have a similar config for the torch model head and think of something with the sklearn model head?

Welcome to any suggestions!

### DIFF
--- a/src/setfit/config.py
+++ b/src/setfit/config.py
@@ -7,13 +7,13 @@ from transformers import PretrainedConfig
 @dataclass
 class SetFitConfig:
     """
-    Config for SetFitModel.
+    Configuration for SetFitModel.
 
-    Parameters:
+    Args:
         model_body (`PretrainedConfig`):
-            Config of the model_body transformer.
+            Configuration of the Sentence Transformer body.
         model_head (`Optional[Any]`):
-            Config of the model_head.
+            Configuration of the head.
     """
 
     model_body: PretrainedConfig


src/setfit/config.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from transformers import PretrainedConfig
+
+
+@dataclass
+class SetFitModelConfig:
```
lewtun, line 8: I think it would be good to align with the `transformers` convention to exclude `Model` from the config name:
```suggestion
class SetFitConfig:
```

src/setfit/config.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Config for SetFitModel.
+
+    Parameters:
```
lewtun, line 12: nit since we try to use `Args` in our docstrings:
```suggestion
    Args:
```

src/setfit/config.py
```diff
@@ -0,0 +1,20 @@
...
+        model_body (`PretrainedConfig`):
+            Config of the model_body transformer.
```
lewtun, line 14: nit:
```suggestion
            Configuration of the Sentence Transformer body.
```

lewtun (commented):
Thanks a lot for this contribution @jegork ! It's something I've also noticed would be quite useful when pushing models to the Hub and being able to store the label mappings for inference.

I think for now I'd like us to come up with a better API for handling the distinction between model bodies and heads before accepting this PR (because it could quickly become obsolete)

The issue in question is this one: https://github.com/huggingface/setfit/issues/238

I'll come back to this PR once we have a clearer vision for how the API will look moving forward

Issue #206, opened 2022-11-28
Moving trained model between devices


jegork:
Hello!
Is there a way to move the trained SetFit model to another device?
E.g. I train a model on a GPU, and after initializing the model on a CPU-only machine I get 
`RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.`

I have managed to fix the error by running:

```python
model.model_body = model.model_body.to("cpu")
model.model_head = model.model_head.to("cpu")
```

However, I guess it depends on the type of the head (in case of the Logistic regression head there is no need to move it).

If there is no single method to move a model between devices, shall I open a PR?

blakechi:
Hi @jegork,

Thanks for raising this issue up!

Ya, I think by simply adding a `to` function for `SetFitModel` can fulfill your needs. But I think there might have something we need to fix due to the error you mentioned. Could you provide more details on how you initializing the model? Or maybe provide a snippet to reproduce the error? Thanks!

jegork:
Hey @blakechi

I've fine-tuned a SetFit model, that was initialized using:
```python
SetFitModel.from_pretrained(
            model_name,
            use_differentiable_head=True,
            head_params={"out_features": 32},
        )
``` 

Afterwards, I've saved it using `.save_pretrained` and loaded into HF hub. Later I've initialized it again using same code above, and it lead me to this error. I fixed it by moving both head and body to CPU and saving and uploading it again to HF hub, which solved the issue.

blakechi:
Thank you for your detailed explanation! I guess the error raised because of the initialization method of SetFitHead. Will validate my guess shortly.

Thanks again for raising this up!

jegork:
@blakechi 
Apparently, there is something with the way that joblib serializes the torch head weights. Because I see that in SetFitModel it is loaded via `joblib.load`, which leads to the error.

However, it should work fine if you load the same file using `torch.load`.
But using `torch.load("model_head.pkl", map_location=torch.device("cpu"))`, I get the same error

jegork:
I think the only way is to either pass the head type (sk/torch) into the config .json or rely on the extension of the head file (.bin for torch and .pkl for sk) and save/load respectively using joblib/torch. I can add a PR this week if we can come to an agreement on how to make it best :)

JackMurphy97:
[...]

PR #229, opened 2022-12-12
Add SetFitModel.to

### LINKED ISSUE #193: picking which GPU to run the model on
Hello,
First, thanks for the great lib and great work. I was wondering if there is a way to pick which cuda device to run my model on ? tried to search for it but it seems to automatically detect and just pick.

jegork:
Fixes #193, #206

Add a simple method to move the body and/or head to the specified device.

### DIFF
--- a/tests/test_modeling.py
+++ b/tests/test_modeling.py
@@ -213,9 +213,7 @@ def test_setfit_from_pretrained_local_model_with_head(tmp_path):
 
 
 def test_to_logistic_head():
-    model = SetFitModel.from_pretrained(
-        "sentence-transformers/paraphrase-albert-small-v2"
-    )
+    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
     devices = (
         [torch.device("cpu"), torch.device("cuda", 0), torch.device("cpu")]
         if torch.cuda.is_available()


tomaarsen (approved):
Note that this does mean that we differ from e.g. torch, which creates a moved copy rather than moving the e.g. tensor itself to a different device. (See the Note in the [`Tensor.to` docs](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html) for details on how torch handles `to`)

That said, I don't mind the difference much.

lewtun (approved):
This is wonderful @jegork - I was actually annoyed by not being able to do this myself recently :)

Thank you for the clean implementation and for adding tests ❤️ !

lewtun:
Hey @jegork we merged some changes to the CI before I merged your PR and it turns out some of the tests fail on `main`. I've reverted your change for now, but would you mind opening a new PR with your contribution once #234 is merged?

Thanks and sorry for the bother!

connormeaton:
I found my way to this page because I am having trouble running a setfit model locally on cpu (I trained on a GPU instance, saved with joblib). I believe this PR attempts to solve this issue. How do I use this code to run my gpu trained model on cpu?

Here is my save code:
```
joblib.dump(trainer, 'model.joblib')
```
Here is my error when running:
```
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.
```

I get the same error when loading both of the ways below:
```
new_trainer = joblib.load('model.joblib')
torch.load('model.joblib', map_location='cpu')
torch.load('model.joblib', map_location=torch.device('cpu'))
```

PR #231, opened 2022-12-13
Prevent overriding the sample size in `sample_dataset`

### LINKED ISSUE #230: num_sample in the sample_dataset function is overwritten
Hello!

`num_sample` argument in `sample_dataset` function does not work normally.
If the data is less than `num_samples` on a particular label, then less than `num_samples` will be sampled from the next label.

I think it's because `num_sample` is overwritten in the following code:
https://github.com/huggingface/setfit/blob/eee595ede962b9a9bbe62d4d919f5629d2fc2868/src/setfit/data.py#L100-L103

how about changing the variable name to the following code:
```python
 for label in range(num_labels): 
     data = shuffled_dataset.filter(lambda example: int(example[label_column]) == label) 
     num_label_samples = min(len(data) // 2, num_samples) 
     samples.append(data.select([i for i in range(num_label_samples)])) 
```

tomaarsen:
Resolves #230

Hello!

## Pull Request Overview
* Prevent `num_samples` from being reduced for all subsequent labels if one label with less than `num_samples` samples is encountered.
* Add test to prevent regression.

## Bug Details
Please see #230 by @solomonno for some details surrounding the bug. As mentioned by Sungjun in that issue, these lines are the culprit:
https://github.com/huggingface/setfit/blob/eee595ede962b9a9bbe62d4d919f5629d2fc2868/src/setfit/data.py#L100-L103

In short, `num_samples` is modified, and may be decremented when we loop over the different labels. I'd rather show you with a code example:

## How to reproduce
```python
import string
from datasets import Dataset
from setfit import sample_dataset

# Generate a dataset where "a" has label 0, and all other letters have label 1
ds = Dataset.from_dict({"text": string.ascii_letters, "label": [0] + 51 * [1]})

train_ds = sample_dataset(ds, num_samples=8)
print(train_ds)
```
This script outputs:
```python
Dataset({
    features: ['text', 'label'],
    num_rows: 0
})
```
Note: 0 rows. This is because `num_samples` is reduced to 0 in the first iteration of the loop.

## After this PR
When running the same script on the branch of this PR, the output becomes:
```python
Dataset({
    features: ['text', 'label'],
    num_rows: 8
})
```

I've made a test that ensures that the num_rows of this script is 8.

## What now?
The output after this PR is *almost* as I would have expected, with the following 2 exceptions:
1. I would expect a `num_rows` of 9: 1 sample for label 0 and 8 samples for label 1.
2. I would expect a warning that for label 0, it was not possible to sample the full `num_samples` samples.

For exception 1, I don't fully understand why we even perform the integer division by 2 on this line:
https://github.com/huggingface/setfit/blob/eee595ede962b9a9bbe62d4d919f5629d2fc2868/src/setfit/data.py#L102
It means that the "second half" of each label in the dataset after shuffling can never be sampled. Why is that? I assume it's unrelated to the contrastive learning.

@lewtun is there a reasoning for this, or should I remove the integer division?
I can also implement the warning from exception 2 (in a different PR?).

Thank you @solomonno for raising the issue and providing the fix. I've added you as a co-author to the commit.

- Tom Aarsen

### DIFF
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -99,7 +99,7 @@ def sample_dataset(dataset: Dataset, label_column: str = "label", num_samples: i
     samples = []
     for label in range(num_labels):
         data = shuffled_dataset.filter(lambda example: int(example[label_column]) == label)
-        num_label_samples = min(len(data) // 2, num_samples)
+        num_label_samples = min(len(data), num_samples)
         samples.append(data.select([i for i in range(num_label_samples)]))
 
     all_samples = concatenate_datasets(samples)
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -126,7 +126,7 @@ def test_sample_dataset_with_label_column():
 def test_sample_dataset_with_unbalanced_ds(unbalanced_dataset):
     num_samples = 8
     ds = sample_dataset(unbalanced_dataset, num_samples=num_samples)
-    # The dataset ought to have just `num_samples` rows, as `unbalanced_dataset`
-    # only has one label for which multiple entries exist. So, we can only sample
-    # for one label, leading to `num_samples` rows in the final dataset.
-    assert ds.num_rows == num_samples
+    # The dataset ought to have just `num_samples + 1` rows, as `unbalanced_dataset`
+    # has one label with more than `num_samples` entries and another label with just 1 row.
+    # We sample `num_samples` from the former, and 1 from the latter.
+    assert ds.num_rows == num_samples + 1


src/setfit/data.py
```diff
@@ -99,8 +99,8 @@ def sample_dataset(dataset: Dataset, label_column: str = "label", num_samples: i
     samples = []
     for label in range(num_labels):
         data = shuffled_dataset.filter(lambda example: int(example[label_column]) == label)
-        num_samples = min(len(data) // 2, num_samples)
-        samples.append(data.select([i for i in range(num_samples)]))
+        num_label_samples = min(len(data) // 2, num_samples)
```
lewtun, line 102: Good catch about the uneccesary integer division - `len(data)` should always be an `int` so this is probably what we want:
```suggestion
        num_label_samples = min(len(data) / 2, num_samples)
```

lewtun (approved):
Thanks a lot for fixing this sneaky bug @tomaarsen!

The suggestion to remove the integer division makes sense too - I was probably coding in a rush and thought it was needed at the time :)

src/setfit/data.py
tomaarsen: Although `len(data)` is always an int, `len(data) / 2` is not, e.g. if `data` has a length of 3. 
That said, isn't the best solution simply `num_label_samples = min(len(data), num_samples)`?

lewtun: Yes, I think you're right! Let's switch to that for now :)

tomaarsen: Done!

tomaarsen:
This should be all set now!

lewtun (approved):
Thanks for iterating @tomaarsen - LGTM 🔥 !

dcruiz01:
This update has yet to be pushed to pypl, correct @lewtun?

tomaarsen:
That is correct, @dcruiz01

PR #234, opened 2022-12-14
Revert "Add SetFitModel.to"

lewtun:
Reverts huggingface/setfit#229 due to failing CI on `main`

cc @tomaarsen


tomaarsen:
@lewtun 

I looked into this briefly. It seems like this is an issue with the Python dependency caching that I introduced in #233, rather than an issue with #229. All of the CI runs that loaded the cache threw an error, and the few runs that had to reinstall the dependencies worked as intended.
It seems like `setfit` itself was included in the cache, meaning that the imports for setfit used the old version rather than the checked-out code.

I'll look into it. To be concrete: #229 is safe and this CI issue does not negatively affect a release.

- Tom Aarsen

PR #235, opened 2022-12-14
Prevent the CI from reusing old cached versions of `setfit`

tomaarsen:
Resolves the issue from https://github.com/huggingface/setfit/pull/234#issuecomment-1351541166

Hello!

## Pull Request Overview
* Regardless of whether the pip dependency cache is used or if the pip dependencies are installed fresh, always install `setfit` anew just prior to running the CI tests.

## Details
#233 improved on the CI testing suite by (among other things) implementing pip dependency caching. However, as was experienced in [this GitHub Action](https://github.com/huggingface/setfit/actions/runs/3695558132), whenever the pip dependency cache was used, older versions of setfit would be used in the newer tests. This caused very confusing test breakages, as new tests were ran with an old version of setfit.

See https://github.com/huggingface/setfit/pull/234#issuecomment-1351541166 for more details.

## The changes
I tried various approaches in a [PR on my fork](https://github.com/tomaarsen/setfit/pull/2), and discovered that the changes from this PR worked. I tested this by first pushing the commit from this PR, and letting the CI build a cache from that. Then, I pushed another commit that adds `raise Exception()` to the `SetFitModel`. As hoped, the test suite immediately picked up on that and failed, indicating that no old cached setfit version was being used, but only the newest version.

Furthermore, I'll `git cherry-pick` e29486f1f99072ed1caf24079799aa4d7aa9e9bd and make a new PR to reintroduce it. It was an unfortunate victim in this bug, as the caching bug made it seem like e29486f1f99072ed1caf24079799aa4d7aa9e9bd was broken, while it was not.

- Tom Aarsen

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,11 +36,14 @@ jobs:
           path: ${{ env.pythonLocation }}
           key: python-dependencies-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.requirements }}-${{ hashFiles('setup.py') }}-${{ env.pythonLocation }}
 
-      - name: Install dependencies on cache miss
+      - name: Install external dependencies on cache miss
         run: |
           python -m pip install --no-cache-dir --upgrade pip
           python -m pip install --no-cache-dir ${{ matrix.requirements }}
         if: steps.restore-cache.outputs.cache-hit != 'true'
+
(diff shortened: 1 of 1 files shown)


lewtun (approved):
Thanks for detailed analysis and for submitting a fix @tomaarsen - this looks good to me!

PR #252, opened 2022-12-29
Add proper model card

### LINKED ISSUE #118: Update model card generation for SetFit tasks
The current `push_to_hub()` implementation uses `sentence-transformers`'s default model card which isn't quite correct because the task is `text-classfication` and not `feature-extraction`.

This results in a model card that is (a) missing nice information about the model / training process and (b) has the wrong widget:

Screen Shot 2022-11-15 at 09 57 48

It would be great to have a dedicated model card for `setfit` models that includes a template with:

* The model name and a code snippet on how to use it
* A pointer back to this repo
* Tags like `setfit` and `text-classification` that we can use to track / filter these models

Now that the `huggingface_hub` library has dedicated APIs for [model cards](https://huggingface.co/docs/huggingface_hub/how-to-model-cards), we could use these to make this process simple

lewtun:
This PR adds a proper model card for SetFit models vs the default one inherited from `sentence-transformers`.

Here's an example of what the end result looks like: https://huggingface.co/lewtun/setfit-new-model-card

Closes #118

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -398,9 +398,7 @@ class SetFitModel(PyTorchModelHubMixin):
 
         return self
 
-    def create_model_card(
-        self, path: str, model_name: Optional[str] = "SetFit Model", train_datasets: List[str] = None
-    ) -> None:
+    def create_model_card(self, path: str, model_name: Optional[str] = "SetFit Model") -> None:
         """Creates and saves a model card for a SetFit model.
 
         Args:


lewtun:
Oops, need to fix those tests!

tomaarsen (approved):
LGTM.

In other HF products, is there a common way for users to be able to specify a different model card template somehow? I haven't used functionality like that myself and I don't know if it exists or not. I'm asking because the approach from this PR can't really be updated by a user who wants to push using a different model card.
That said, users can always manually update and commit the README from a cloned version of the model repository, or just edit the card in the UI on the Hub website, so it's not a big deal. I'm just wondering as it's preferable to align functionalities with other HF products where possible.

lewtun:
> In other HF products, is there a common way for users to be able to specify a different model card template somehow?

Since v0.10 of the `huggingface_hub` library, we do have a programatic way to create model cards from string templates or even Jinja ([docs](https://huggingface.co/docs/huggingface_hub/v0.11.0/en/how-to-model-cards)). 

However, I'm a bit reluctant to expose these features right now because it's a bit at odds with how our other libraries like `transformers` and `diffusers` create model cards - in most cases, it's just a generic template and users can modify them as needed.

See for instance the models generated from `transformers.Trainer`: https://huggingface.co/models?library=transformers&other=generated_from_trainer

tomaarsen:
I think we should just stick with this approach, then. Its simplicity and ease of maintenance has value, too.

PR #257, opened 2023-01-06
Add has_differentiable_head property to SetFitModel

zachschillaci27:
Adding a `has_differentiable_head` property to `SetFitModel` to simplify the repeated `isinstance(self.model_head, nn.Module)` checks.

Please let me know if any changes are needed!

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -3,7 +3,6 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
 import evaluate
 import numpy as np
-import torch
 from sentence_transformers import InputExample, losses
 from sentence_transformers.datasets import SentenceLabelDataset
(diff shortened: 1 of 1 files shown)


zachschillaci27:
> Seems good! This will certainly do for the time being. I'm not sure whether it's necessary for `has_differentiable_head` to be a property, as a simple class attribute in the `__init__` likely also works fine. Presumably the value for `has_differentiable_head` never changes over time, after all.
> 
> Thanks @zachschillaci27

Thanks for merging!

I agree with your point. I chose a property mainly to prevent the attribute from being accidentally modified and to cover the strange (likely discouraged!) case of a user modifying the `model_head` of a `SetFitModel` instance.

tomaarsen:
Indeed! That would be a really difficult to spot bug. That is why I believe this PR was ready to merge as it stood.

Perhaps the best solution is to use a `has_differentiable_head` attribute and convert `model_head` into a property with a setter method that updates the `has_differentiable_head` attribute. It's a little bit overengineered for a simple problem, though.

PR #262, opened 2023-01-09
Automatically create summary table after `scripts/setfit/run_fewshot.py`

### LINKED ISSUE #246: Automatically execute `scripts/create_summary_table.py` to reproduce average scores from paper
The script in `scripts/create_summary_table.py` must be used to reproduce SetFit / baseline results from the paper. It calculates per-dataset average over the 10 training splits and saves the summary table to a file. However, this is scripts is not mentioned anywhere in the repo. It will be nice if it can be run automatically at the end of each SetFit / baseline run.

tomaarsen:
Closes #246

Hello!

## Pull Request overview
* Automatically create summary table after `scripts/setfit/run_fewshot.py`.
* Resolve potential NameError in `create_summary_table` if the `sample_sizes` is an empty list.

## Details
Note that [`create_summary_table.py`](https://github.com/huggingface/setfit/blob/main/scripts/create_summary_table.py) is located one directory higher than [`run_fewshot.py`](https://github.com/huggingface/setfit/blob/main/scripts/setfit/run_fewshot.py). As a result, if we want to preserve that `run_fewshot.py` can be called like `python run_fewshot.py` from the `scripts/setfit` directory, then we must apply the following hack:
```python
sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
from create_summary_table import create_summary_table
```

As opposed to the "normal" approach of:
```python
from scripts.create_summary_table import create_summary_table
```
which would only work from the repository's root directory. All of these work due to the hack:
```
[path]/setfit/scripts/setfit> python run_fewshot.py
[path]/setfit/scripts> python setfit/run_fewshot.py
[path]/setfit> python scripts/setfit/run_fewshot.py
```

---

Furthermore, `metric_name` was only defined in a loop body, and then returned after the loop. If the list over which was being looped was empty, then this would cause a `NameError`, as `metric_name` was never defined. I've avoided that by also initializing the variable before the loop.

cc: @danielkorat 

- Tom Aarsen

### DIFF
--- a/scripts/setfit/run_fewshot.py
+++ b/scripts/setfit/run_fewshot.py
@@ -173,7 +173,8 @@ def main():
                     sort_keys=True,
                 )
 
-    # Create a summary_table.csv file from all of the results in `output_path`
+    # Create a summary_table.csv file that computes means and standard deviations
+    # for all of the results in `output_path`.
     create_summary_table(str(output_path))
 
 


danielkorat:
Thanks @tomaarsen !
What's the status of this PR?

tomaarsen:
It's awaiting review/ready to merge. I tend not to merge my own PRs without someone's review.

tomaarsen:
That said, if you're willing and able to provide a review, I'd appreciate it :)

danielkorat:
@tomaarsen Sure, it was on my to do list :)

danielkorat (approved):
@tomaarsen Looks very good, thanks! Feel free to merge when possible.

tomaarsen:
Wonderful! Thanks for the review.

PR #265, opened 2023-01-11
Refactor to introduce `Trainer` & `TrainingArguments`, add SetFit ABSA

### LINKED ISSUE #179: Refactor training hyperparameters in `SetFitTrainer` to dedicated class

### LINKED ISSUE #238: Discussion on refactoring
Hello!

# Why might we want to refactor?

I've opened this issue is a response to several issues that SetFit is currently tackling:
1. #179, i.e. the desire to implement a `TrainingArguments` à la 🤗 Transformers and PyTorch Lightning.
2. Difficulties with different heads as experienced (among other places) in #207 and #212. To be concrete, this project relies on `if isinstance(..., torch.Tensor)`, `if isinstance(..., nn.Module)`, `if isinstance(..., LogisticRegression)`, etc. way too much.
3. With the introduction of the differentiable head, training is no longer simply `trainer.train()`: to use the differentiable head correctly, you must `trainer.freeze()` the head, then `trainer.train()` for the body, then `trainer.unfreeze(...)` the head (but maybe keep the body frozen?), and then `trainer.train(...)` for the head, with a whole host of different parameters.
4. `use_differentiable_head=True` is too vague, not well enough documented, and not flexible enough if we want to ever have support for more than 2 heads.
5. Passing parameters for the head is also unintuitive, as it currently needs to be done like `head_params={"out_features": num_classes}`. 

A natural consequence of issue 2 is that implementing a new head is quite problematic: it requires editing many different files, and may introduce bugs in other contexts (e.g. when using different heads). One of the major headaches is a type mistmatch between the body output and the head input, which has already caused bugs (#237, #207).

Issue 3 is quite noteworthy, as it shows us that different we may want to use different hyperparameters for step 1 (embedding finetuning) than for step 2 (classifier training) in the SetFit training process.

# Goals

With a refactor, I hope to achieve several primary goals:
1. Ensure a familiar, intuitive training scheme involving `Trainer` and `TrainingArguments` classes.
2. Additionally, ensure that the model can also be trained *without* a `TrainingArguments` class. In short, a fit method of a component (body or head) must not accept a `TrainingArguments` instance as a parameter, but merely the hyperparameters like `learning_rate`, etc.
3. Ensure that the entire SetFit model can be trained with a single `trainer.train()`, regardless of the head chosen.
4. Ensure separation of issues by requiring that each of the heads are moved into separate files and implement a fixed interface. This interface requires the use of `torch.Tensor` to avoid type issues.
6. Introduce a `head` parameter to `SetFitModel.from_pretrained` where users can provide implementations of the aforementioned head interface. This includes our provided logistic regression and linear heads, but also custom heads that users can come up with. 

# Structure
I imagine one of two structures: a simple one, with an interface only for the SetFit head, and a more complex one, where a fixed interface is introduced for the body, too. 

<details><summary>Simple structure</summary>

```
└───setfit
    │   data.py
    │   integrations.py
    │   logging.py
    │   losses.py
    │   pipeline.py
    │   trainer.py
    │   trainer_distillation.py
    │   training_args.py
    │   utils.py
    │   __init__.py
    │
    ├───components
    │   │   modeling.py  # Contains SetFitModel
    │   │
    │   ├───body
    │   │   model.py
    │   │   training_args.py
    │   │   utils.py
    │   │
    │   └───head
    │       │   api.py
    │       │   training_args.py
    │       │
    │       ├───linear
    │       │       model.py
    │       │       training_args.py
    │       │
    │       ├───logistic
    │       │       model.py
    │       │       training_args.py
    │       │
    │       └───sklearn
    │               model.py
    │               training_args.py
    │
    └───exporters
            onnx.py
            utils.py
            __init__.py
```
</details>

<details><summary>Complex structure</summary>

```
└───setfit
    │   data.py
    │   integrations.py
    │   logging.py
    │   losses.py
    │   pipeline.py
    │   trainer.py
    │   trainer_distillation.py
    │   training_args.py
    │   utils.py
    │   __init__.py
    │
    ├───components
    │   │   api.py  # Contains SetFitI, the superclass of SetFitHeadI and SetFitBodyI
    │   │   modeling.py  # Contains SetFitModel
    │   │
    │   ├───body
    │   │   │   api.py  # Contains SetFitBodyI(SetFitI)
    │   │   │
    │   │   └───sentence_transformer
    │   │           model.py  # Contains SetFitSTBody(SetFitBodyI)
    │   │           training_args.py  # Provides different hyperparameter defaults for SetFitSTBody
    │   │           utils.py
    │   │
    │   └───head
    │       │   api.py  # Contains SetFitHeadI(SetFitI)
    │       │
    │       ├───linear
    │       │       model.py  # Contains SetFitLinearHead(SetFitHeadI)
    │       │       training_args.py  # Provides different hyperparameter defaults for SetFitLinearHead
    │       │
    │       ├───logistic
    │       │       model.py  # Contains SetFitLogisticHead(SetFitSklearnHead)
    │       │       training_args.py  # Provides different hyperparameter defaults for SetFitLogisticHead
    │       │
    │       └───sklearn
    │               model.py  # Contains SetFitSklearnHead(SetFitHeadI)
    │               training_args.py  # Provides different hyperparameter defaults for SetFitSklearnHead
    │
    └───exporters
            onnx.py
            utils.py
            __init__.py
```
</details>

## Training Arguments
Both of these structures have files for `training_args.py`, which contains a dataclass with various hyperparameters. Each of the more nested `training_args.py` files have subclasses of the `TrainingArguments` where defaults are overridden to defaults that would work appropriately for that component and training phase. The `TrainingArguments` class may look like so:
```python
from __future__ import annotations

from abc import ABC, abstractclassmethod
from dataclasses import dataclass, fields


@dataclass
class TrainingArguments(ABC):
    learning_rate: float
    ...

    @abstractclassmethod
    def embeddings_default(cls) -> TrainingArguments:
        raise NotImplementedError()

    @abstractclassmethod
    def classifier_default(cls) -> TrainingArguments:
        raise NotImplementedError()

    def to_dict(self):
        # filter out fields that are defined as field(init=False)
        return {field.name: getattr(self, field.name) for field in fields(self) if field.init}
```

Crucially, we may need up to 3 different `TrainingArgument` instances for the head and body in all situations in order to deal with the issue that training arguments for the body may differ for step 1 and step 2 of training SetFit:
```python
trainer = Trainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    embedding_body_args=TrainingArguments(...),
    classifier_body_args=TrainingArguments(...),
    classifier_head_args=TrainingArguments(...)
)
trainer.train()
```
Note that each of these arguments are optional. If one of the training arguments is not supplied, then we can use the default arguments for each of the components (body, head) for that training step (step 1: embeddings, step 2: classifier) to get the right hyperparameters. This is best described with an example of the new proposed `Trainer.train` method:
<details><summary>Trainer.train method</summary>

```python
    def train(
        self,
        embedding_body_args: Optional[BodyTrainingArguments] = None,
        classifier_body_args: Optional[BodyTrainingArguments] = None,
        classifier_head_args: Optional[HeadTrainingArguments] = None,
    ):
        # Simple example excluding column_mapping, hyperparameter tuning trials, etc.
        x_train: List[str] = train_dataset["text"]
        y_train: List[int] = train_dataset["label"]

        self.train_embeddings(x_train, y_train, embedding_body_args)
        self.train_classifier(x_train, y_train, classifier_body_args, classifier_head_args)

    def train_embeddings(
        self, x_train: List[str], y_train: List[int], embedding_body_args: Optional[BodyTrainingArguments] = None
    ):
        embedding_body_args = (
            embedding_body_args or
            self.embedding_body_args or
            self.model.body.args_class.embeddings_default()
        )

        self.model.body.fit(x_train, y_train, **embedding_body_args.to_dict())

    def train_classifier(
        self,
        x_train: List[str],
        y_train: List[int],
        classifier_body_args: Optional[BodyTrainingArguments] = None,
        classifier_head_args: Optional[HeadTrainingArguments] = None,
    ):
        classifier_body_args = (
            classifier_body_args or
            self.classifier_body_args or
            self.model.body.args_class.classifier_default()
        )
        classifier_head_args = (
            classifier_head_args or
            self.classifier_head_args or
            self.model.head.args_class.classifier_default()
        )
        # If we don't train end-to-end, then simply get the embeddings of the training input,
        # and fit the head using it
        if not classifier_head_args.end_to_end:
            embeddings = self.model.encode(x_train)
            self.model.head.fit(embeddings, y_train)
        else:
            # Otherwise, fit the model itself, which implement a PyTorch end-to-end training loop.
            # The body_learning_rate is passed from the classifier body arguments, and for the rest
            # of the parameters we use the classifier head arguments.
            self.model.fit(
                x_train,
                y_train,
                body_learning_rate=classifier_body_args.learning_rate,
                **classifier_head_args.to_dict(),
            )
```
</details>
Note here that the training arguments are provided by unpacking a dictionary representation of the hyperparameters.

## Head Interface
To tackle issue 2, I propose an interface which all SetFit heads must follow. A potential implementation might be:

<details><summary>Example head interface (or abstract class, rather)</summary>

```python
from __future__ import annotations

from abc import ABC, abstractmethod, abstractproperty
from dataclasses import dataclass
from typing import List, Optional, Union, Iterator

from torch import nn
import torch

from setfit.training_args import TrainingArguments


@dataclass
class SetFitHeadI(ABC):
    @abstractproperty
    def args_class(self) -> TrainingArguments:
        pass

    @abstractproperty
    def device(self) -> Optional[Union[str, torch.device]]:
        pass

    @abstractmethod
    def fit(self, x_train: List[str], y_train: List[int], **kwargs) -> None:
        pass

    @abstractmethod
    def to(self, device: Union[str, torch.device]) -> SetFitHeadI:
        pass

    @abstractmethod
    def train(self, mode: bool = True) -> None:
        pass

    @abstractmethod
    def parameters(self) -> Iterator[nn.Parameter]:
        pass

    @abstractmethod
    def predict(self, inputs: torch.Tensor) -> torch.Tensor:
        pass

    @abstractmethod
    def predict_proba(self, inputs: torch.Tensor) -> torch.Tensor:
        pass
```
</details>
Note that the inputs and outputs for `predict` and `predict_proba` are `torch.Tensor`s to prevent type issues. Then, an implementation of this interface may be:

<details><summary>Example head implementation</summary>

```python
from __future__ import annotations

from dataclasses import dataclass, field
from typing import Iterator, List, Optional, Union

import numpy as np
from torch import nn
import torch
from sklearn.linear_model import LogisticRegression
from sklearn.multiclass import OneVsRestClassifier
from sklearn.multioutput import ClassifierChain, MultiOutputClassifier

from setfit.components.head.sklearn.training_args import SklearnHeadTrainingArguments

from ..api import SetFitHeadI


@dataclass
class SetFitSklearnHead(SetFitHeadI):
    clf: LogisticRegression = None
    multi_target_strategy: str = None

    def __post_init__(self):
        if self.multi_target_strategy is not None:
            multi_target_mapping = {
                "one-vs-rest": OneVsRestClassifier,
                "multi-output": MultiOutputClassifier,
                "classifier-chain": ClassifierChain,
            }
            if self.multi_target_strategy in multi_target_mapping:
                self.clf = multi_target_mapping[self.multi_target_strategy](self.clf)
            else:
                raise ValueError(
                    f"multi_target_strategy {self.multi_target_strategy!r} is not supported. "
                    f"Choose one from {list(multi_target_mapping.keys())!r}."
                )

    def to(self, device: Union[str, torch.device]) -> SetFitSklearnHead:
        return self

    @property
    def args_class(self) -> SklearnHeadTrainingArguments:
        return SklearnHeadTrainingArguments

    def fit(self, embeddings: torch.Tensor, y_train: List[int]) -> SetFitSklearnHead:
        embeddings = embeddings.detach().cpu().numpy()
        self.clf.fit(embeddings, y_train)
        return self

    def predict(self, inputs: torch.Tensor) -> torch.Tensor:
        X = inputs.detach().cpu().numpy()
        y_pred: np.ndarray = self.clf.predict(X)
        return torch.Tensor(y_pred)

    def predict_proba(self, inputs: torch.Tensor) -> torch.Tensor:
        X = inputs.detach().cpu().numpy()
        y_pred: np.ndarray = self.clf.predict_proba(X)
        return torch.Tensor(y_pred)

    @property
    def device(self) -> Optional[Union[str, torch.device]]:
        return None

    def parameters(self) -> Iterator[nn.Parameter]:
        return []

    def train(self, mode: bool = True) -> None:
        pass


@dataclass
class SetFitLogisticHead(SetFitSklearnHead):
    clf: LogisticRegression = field(default_factory=LogisticRegression)
```
</details>

The `SetFitSklearnHead` class wraps around a Sklearn classifier, e.g. `LogisticRegression`. Additionally, the `SetFitLogisticHead` is a simple subclass of the aforementioned sklearn wrapper class, and it sets the classifier to be a `LogisticRegression` by default.

# Where do I stand?
As you may suggest based on the detailed examples, I have started work on this refactor. In particular, I have set up the very first steps of the implementation in my [refactor branch](https://github.com/tomaarsen/setfit/tree/refactor). It's a working MVP, but it's still far from polished. Note that I have so far attempted to implement the complex structure described above.

Using this branch, the following snippet works correctly, regardless of the head that is chosen:
<details><summary>SetFit Demo</summary>

```python
from sklearn.linear_model import LogisticRegression

from setfit import SetFitModel, Trainer, sample_dataset
from setfit.components.head.linear.model import SetFitLinearHead
from setfit.components.head.logistic.model import SetFitLogisticHead
from setfit.training_args import TrainingArguments


# Load a SetFit model from Hub using whatever head we want.
head = SetFitLogisticHead(clf=LogisticRegression(solver="liblinear"))
# head = SetFitLogisticHead()
# head = SetFitLinearHead()
model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2", head=head)

# Create trainer
trainer = Trainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    # embedding_body_args=TrainingArguments(),
    # classifier_body_args=TrainingArguments(),
    # classifier_head_args=TrainingArguments(),
    column_mapping={"sentence": "text", "label": "label"},
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()

print(metrics)
```
</details>

Both the `SetFitLogisticHead` and `SetFitLinearHead` are implemented. We can pass head parameters directly to the initializations of these heads. If no `head` parameter is used in `SetFitModel.from_pretrained`, then it simply initialises a `SetFitLogisticHead`. Furthermore, `TrainingArguments` can be passed using any of the `embedding_body_args`, `classifier_body_args` and `classifier_head_args` parameters.

So, where do we stand now? Let's refer back to our original issues:
1. `Trainer` and `TrainingArguments` are implemented. However, it is uncommon to have to initialize several of them.
2. Our heads have been split into separate files.
3. Only one `trainer.train()` is needed, even if we use `SetFitLinearHead`. It should be easy to add `freeze_body` and `freeze_head` parameters to the `TrainingArguments`.
4. `use_differentiable_head=True` was replaced by the more flexible `head` parameter.
5. Head arguments are now applied by initializing a head before calling `SetFitModel.from_pretrained`, and no longer need to be passed like `head_params={"out_features": num_classes}`.

You'll find that my primary goals are satisfied with this approach, too.
However, some issues remain:
1. Having `embedding_body_args`, `classifier_body_args` *and* `classifier_head_args` parameters may be difficult to understand without knowledge about SetFit's model structure.
2. Having  `embedding_body_args`, `classifier_body_args` and `classifier_head_args` parameters may be confusing in certain situations, for example, the `classifier_body_args` would never be used if the head is a simple logistic regression head. Perhaps we can also move towards having only `embedding_args` and `classifier_args`? This would already simplify things a lot.
3. Some very important parameters such as `multi_target` have to be specified somewhere, preferably when the model is initialized. Perhaps a solution is to add it as a parameter to `SetFitModel.__init__` and `SetFitModel.from_pretrained`?
4. The end-to-end training from the linear head needs to be implemented on the level of `SetFitModel` rather than `SetFitLinearHead`, because it involves both the head and the body. This may be limited or problematic especially if we opt for the complex structure.
5. I have yet to add support for the trainer distillation, and important features like hyperparameter tuning, exporting, saving and loading models have also not been implemented.
6. There are several open PRs that might be best suited to be included in a larger refactoring like this, such as #187, #203, #207 & #212. I haven't absorbed these into the refactor, with exception of #203.

## What do I want from you reading this?
This discussion is partially to share my progress, partially to look for any form of feedback, and partially to discuss the remaining issues that I last enumerated. More eyes on a problem may help create better solutions. Most crucial of all is feedback on the eventual user experience, which is the main thing that I intend to improve on. With other words, whether you like the updated approach from the last demo snippet.

I'm also considering switching from the complex structure to the standard structure, as it simplifies some things here and there. The complex structure only gives benefits if we would actually ever implement a different body, while it introduces extra complexity from a development and maintenance perspective.

I'll slowly continue working on this monday onwards, and hope that I'll be able to fully implement it to my satisfaction.

cc: @lewtun
@blakechi you may be interested, too :) 

- Tom Aarsen

### LINKED ISSUE #320: Distillation Trainer requires labels in the data
I tried to run Distillation trainer setup as shown below:

```
from setfit import SetFitModel, SetFitTrainer, DistillationSetFitTrainer

train_dataset_student = agnews["train"].shuffle(seed=0).select(range(1000))

#Distill
student_model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-MiniLM-L3-v2")
teacher_model = teacher_trainer.model
student_trainer = DistillationSetFitTrainer(teacher_model=teacher_model, train_dataset=train_dataset_student, student_model=student_model, eval_dataset=test_dataset, loss_class=CosineSimilarityLoss, metric="accuracy", batch_size=16, num_iterations=20, num_epochs=1)
```

I am getting this error

```
ValueError: A column mapping must be provided when the dataset does not contain the following columns: {'text', 'label'}
```

My dataset has only one column called "text"
```
Dataset({
    features: ['text'],
    num_rows: 1000
})
```

The whole point of distillation is not having labelled data. I think the distillation trainer does training for the distilled student body and teacher logits at the end and thus requires labels for that, but how can we provide those few labels separately?

tomaarsen:
Closes #179, closes #238, closes #320

Hello!

## Pull Request overview
* Additions:
  * `Trainer`, a modified version of the old `SetFitTrainer`.
  * `DistilledTrainer`, a modified version of the old `DistillationSetFitTrainer`.
  * `TrainingArguments`, a dataclass used throughout the updated `Trainer` classes.
* Changes & Deprecations:
  * `SetFitTrainer` and `DistillationSetFitTrainer`: a soft deprecation, i.e. old code should give a warning, but still works.
  * `Trainer.unfreeze` and `Trainer.freeze` now simply point to `SetFitModel.unfreeze` and `SetFitModel.freeze`. This involves a soft deprecation for the `keep_body_frozen` parameter, i.e. old code should give a warning, but still works. Furthermore, `Trainer.freeze()` now freezes the full model, not just the head.
  * `Trainer.train` no longer accepts training arguments like `num_epochs`: a fairly soft deprecation, i.e. old code no longer works like before as the keyword arguments are ignored, but the code won't crash. A warning is thrown instead. This will likely affect all users of the differentiable head.
  * `SupConLoss` was moved to `src/setfit/losses.py`.
* Hard removals:
  * `SetFitBaseModel` and `SKLearnWrapper`, which were both unused.
* Tests:
  * Introduced `test_deprecated_trainer_distillation.py` and `test_deprecated_trainer.py`. These files contain old unmodified tests, and they show that despite drastic changes, old code still generally works (with exception of the aforementioned `Trainer.train` training argument deprecation).
  * Updated `test_trainer_distillation.py` and `test_trainer.py` to use the new training procedure.
  * Introduced `test_training_args.py`.

## Goals
Based on my own experiments and useful feedback from @kgourgou in the refactoring discussion in #238, I have reduced my goal for this refactoring to the following:
1. Ensure a familiar, intuitive training scheme involving `Trainer` and `TrainingArguments` classes.
2. Additionally, ensure that the model can also be trained *without* a `TrainingArguments` class. In short, a fit method of a component (body or head) must not accept a `TrainingArguments` instance as a parameter, but merely the hyperparameters like `learning_rate`, etc.
3. Ensure that the entire SetFit model can be trained with a single `trainer.train()`, regardless of the head chosen.
4. Ensure a smooth and unintrusive way for users to update their code to after this refactor.

Notably, I'm **not** introducing an interface for classifier heads like I had originally planned, and I'm **not** deprecating `SetFitModel.from_pretrained(..., use_differentiable_head=True)` with `SetFitModel.from_pretrained(..., head=...)`. Perhaps some other time we can have a new discussion about those.

## Upgrading guide
To update your code to work with this PR, the following changes must be made:
1. Replace all uses of `SetFitTrainer` with `Trainer`, and all uses of `DistillationSetFitTrainer` with `DistillationTrainer`.
2. Remove `num_iterations`, `num_epochs`, `learning_rate`, `batch_size`, `seed`, `use_amp`, `warmup_proportion`, `distance_metric`, `margin`, `samples_per_label` from a `Trainer` initialisation, and move them to a `TrainerArguments` initialisation instead. This instance should then be passed to the trainer via the `args` argument.
3. Refactor multiple `trainer.train()`, `trainer.freeze()` and `trainer.unfreeze()` calls that were previously necessary to train the differentiable head into just one `trainer.train()` call by setting `batch_size` and `num_epochs` on the `TrainingArguments` dataclass with Tuples. The first value is for training the embeddings, and the second is for training the classifier. 

I applied this guide to all of the tests to upgrade them to work with the new classes.

An example using a differentiable head:
#### Before:
```python
    use_differentiable_head=True,
    head_params={"out_features": 2},
)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    learning_rate=2e-5,
    batch_size=16,
    num_iterations=20,
    num_epochs=1,
)

trainer.freeze() # Freeze the head
trainer.train() # Train only the body

# Unfreeze the head and unfreeze the body -> end-to-end training
trainer.unfreeze(keep_body_frozen=False)

trainer.train(
    num_epochs=16,
    batch_size=2,
    body_learning_rate=1e-5,
    learning_rate=1e-2,
)
metrics = trainer.evaluate()
```

#### After:
```python
    use_differentiable_head=True,
    head_params={"out_features": 2},
)

# Create Training Arguments
args = TrainingArguments(
    # When an argument is a tuple, the first value is for training the embeddings,
    # and the latter is for training the differentiable classification head:
    batch_size=(16, 2),
    num_iterations=20,
    num_epochs=(1, 16),
    body_learning_rate=(2e-5, 1e-5),
    head_learning_rate=1e-2,
    end_to_end=True,
    loss=CosineSimilarityLoss,
)

# Create Trainer
trainer = Trainer(
    model=model,
    args=args,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()
```

### `SetFitTrainer` & `DistillationSetFitTrainer`
This is a soft deprecation, i.e. these can still be used like before. For reference, the following snippet *still works*.
<details><summary>First snippet from the README, using SetFitTrainer</summary>

```python

)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=4,
    num_iterations=20,  # The number of text pairs to generate for contrastive learning
    num_epochs=1,  # The number of epochs to use for contrastive learning
    column_mapping={"sentence": "text", "label": "label"},  # Map dataset columns to text/label expected by trainer
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()
```
</details>
<details><summary>The corresponding output</summary>

```
Found cached dataset sst2 ([sic])
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:00<00:00, 299.96it/s] 
fb86.arrow
Loading cached processed dataset at [sic]
Loading cached processed dataset at [sic]
Loading cached shuffled indices for dataset at [sic]
[sic]\demo_readme.py:20: DeprecationWarning: `SetFitTrainer` has been deprecated. Please use `Trainer` instead.
  trainer = SetFitTrainer(
Applying column mapping to training dataset
***** Running training *****
  Num examples = 640
  Num epochs = 1
  Total optimization steps = 160
  Total train batch size = 4
Iteration: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 160/160 [02:13<00:00,  1.20it/s] 
Epoch: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [02:13<00:00, 133.68s/it] 
Applying column mapping to evaluation dataset
***** Running evaluation *****
{'accuracy': 0.8818807339449541}
```
</details>
Note that the output contains a DeprecationWarning.

### `keep_body_frozen` on `trainer.unfreeze`:
This is also a soft deprecation, so we can use this like before. Any use of `keep_body_frozen` results in a warning, and `trainer.unfreeze(keep_body_frozen=True)` is made equivalent to `model.unfreeze("head")`, while `trainer.unfreeze(keep_body_frozen=False)` is like `model.unfreeze()`. However, there should be no need for this method anymore, so we may be better off to deprecate it fully.

Please note the following snippet:
```python
trainer.freeze()  # Freeze the head
trainer.train()  # Train the head

# unfreeze and train the body
```
This snippet was common when training with a differentiable head, but stops working after this PR, as `trainer.freeze()` will now fully freeze the model, after which `trainer.train()` fails as it can't perform a `backward` on the frozen sentence transformer body. This causes an error:
```
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

### `Trainer.train` arguments
This is a fairly soft deprecation, so code with e.g. `trainer.train(num_epochs=..., ...)` no longer works like before. The arguments passed via keywords are simply ignored now, while a warning is thrown:

```
[sic]\demo_readme_diff.py:44: DeprecationWarning: `SetFitTrainer.train` does not accept keyword arguments anymore. Please provide training arguments via a `TrainingArguments` instance to the `SetFitTrainer`initialisation or the `SetFitTrainer.train` method.
  trainer.train(
```

## Training Arguments
Note that the `TrainingArgument` class must be able to hold separate values for training the embeddings vs the classifier. Most difficult is the case of learning rates, for which 3 separate values must exist:
1. The learning rate of the sentence transformer body while training the embeddings.
2. The learning rate of the sentence transformer body while training the classifier.
3. The learning rate of the head while training the classifier.

For several other arguments (e.g. `num_epochs`, `batch_sizes`), two values must exist:
1. For training the embeddings.
2. For training the classifier.

For these last two cases, `TrainingArguments(num_epochs=12)` can be used, which will set both `embedding_num_epochs` and `classifier_num_epochs` to 12. Alternatively, `TrainingArguments(num_epochs=(1, 12))` will set `embedding_num_epochs` to 1 and `classifier_num_epochs` to 12. Beyond that, `TrainingArguments(embedding_num_epochs=1, classifier_num_epochs=12)` is also allowed, i.e. they can be set directly. In practice, only `embedding_num_epochs` and `classifier_num_epochs` are used, while `num_epochs` exists only to make setting the arguments more natural.

The learning rate case is similar. For these, `body_learning_rate` and `head_learning_rate` must be set explicitly. Like before, `body_learning_rate` can be either set as a float or as a tuple of floats. If a tuple is given, then the first value is once again for training the embeddings, while the second value is for training the classifier. If instead only one value is given to `body_learning_rate`, then it is internally expanded to a tuple. Afterwards, `body_embedding_learning_rate` and `body_classifier_learning_rate` are either explicitly provided or implicitly set via `body_learning_rate`. Like before, only `body_embedding_learning_rate` and `body_classifier_learning_rate` are used, while `body_learning_rate` exists only to make setting the arguments more natural.

Beyond that, there is a new training argument called `end_to_end`. If the model is initialized with a differentiable head, then this argument specifies whether the sentence transformer body is frozen (if `end_to_end=False`) or not (if `end_to_end` is True) during training of the classifier. This replaces the previous need of `trainer.freeze()` before and during `trainer.train()` calls.

See the Upgrading Guide for an example of how the TrainingArguments may be initialized in a relatively complex case.

## Tests
The new `tests\test_deprecated_trainer.py` and `tests\test_deprecated_trainer_distillation.py` files are equivalent to the [`tests\test_trainer.py`](https://github.com/huggingface/setfit/blob/ac958ae064c215839a36ca1f10070101d0d8c3e2/tests/test_trainer.py) and [`tests\test_trainer_distillation.py`](https://github.com/huggingface/setfit/blob/ac958ae064c215839a36ca1f10070101d0d8c3e2/tests/test_trainer_distillation.py) files at ac958ae064c215839a36ca1f10070101d0d8c3e2, with four exception: In `tests\test_deprecated_trainer.py`, three tests are skipped:
* Two are skipped because they contain `trainer.train(max_length=max_length, ...)` and then expect a warning based on that. However, these arguments are now ignored, so no warning is ever thrown. 
* Another test is skipped because it contains the following deprecated snippet:
   ```python
   trainer.freeze()  # Freeze the head
   trainer.train()  # Train the head

   # unfreeze and train the body
   ```
   See the Deprecation section for more details on why this crashes.

And as a fourth change, `from setfit.modeling import SupConLoss` was changed to `from setfit.losses import SupConLoss`.

Using these tests, it is clear that users will not immediately face dozens of errors when updating. Instead, they may face several warnings and perhaps an error if they are using differentiable heads. 

## Discussion points
1. How should users be notified on how to update their code? A section in the README? Via descriptive deprecation warnings? A link to this PR in the release notes?
2. What should the deprecation timeline be? i.e. when do we remove `SetFitTrainer` altogether? We should include this in the deprecation warning messages.
3. Should we simply deprecate `Trainer.freeze()` and `Trainer.unfreeze`? They should not be needed anymore, and they are simply equivalent to calling `trainer.model.freeze()` and `trainer.model.unfreeze()`. I'm in favour of the deprecation.
4. If we do not deprecate `Trainer.freeze()`, should we then introduce a special error message when people use `trainer.freeze()` followed by `trainer.train()`, which now breaks whereas it worked previously? (See the Deprecation section)
5. Are we a fan of the name `end_to_end` for training a differentiable model in full, i.e. not freezing the body when training the classifier?
6. Should other arguments to `Trainer` be absorbed into `TrainingArguments`? In particular, `loss_class`?
7. Which release should this fall under? 0.6.0? 0.7.0? 1.0.0?

And perhaps most importantly, I'm interested in all of your feedback on how you feel about these changes.

## To do:
There are several things that still need to be done before this can be merged:
* [x] Docstrings for all arguments in `TrainingArgument`.
* [x] Update all method docstrings.
* [x] Pick deprecation timeline for deprecated classes and methods.
* [x] Create additional tests to ensure that e.g. `embedding_batch_size` and `classifier_batch_size` are set correctly whenever `batch_size` is set. 

I'll make a separate PR for updating the README/Notebook/scripts.

Feel free to ask me anything if you have any questions.

cc: @lewtun 

- Tom Aarsen

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -5,9 +5,12 @@ on:
     branches:
       - main
       - v*-release
+      - v*-pre
   pull_request:
     branches:
       - main
+      - v*-pre
+  workflow_dispatch:
 
 jobs:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,9 +5,12 @@ on:
     branches:
       - main
       - v*-release
+      - v*-pre
   pull_request:
     branches:
       - main
+      - v*-pre
+  workflow_dispatch:
 
 jobs:
 


tomaarsen:
Merged 4cad62c491641c6c0b63902674d05af4f10b4bbc...174eb00b18cb6ce46cf5840f083f8f0bc621cee5 into this PR via 94106ccadacf569f4eff7b6073ceb968ed6766b0.
Merged 29c034818131383e0ff9e70cbbb819f15f46da11 into this PR via a39e772a9d26ade84ee4f9674f2d889da67b13df.

tomaarsen:
Merged 0cb8ffdb59f896709a565a4c01bbc4636e7e3b39 into this PR via 7d4ad009581a74a57d8d0921dfcb3d17b4a2b212.

src/setfit/trainer.py
```diff
@@ -226,172 +206,126 @@ def call_model_init(self, params: Optional[Dict[str, Any]] = None):
...
-        l2_weight: Optional[float] = None,
-        max_length: Optional[int] = None,
+        args: Optional[TrainingArguments] = None,
         trial: Optional[Union["optuna.Trial", Dict[str, Any]]] = None,
-        show_progress_bar: bool = True,
+        **kwargs,
     ):
-        """
-        Main training entry point.
+        if kwargs:
```
lewtun, line 223: Turbo nit:
```suggestion
        if kwargs is not None:
```

src/setfit/trainer.py
```diff
@@ -226,172 +206,126 @@ def call_model_init(self, params: Optional[Dict[str, Any]] = None):
...
+            warnings.warn(
+                f"`{self.__class__.__name__}.train` does not accept keyword arguments anymore. "
```
lewtun, line 225: Do you think we should put a deprecation warning for `v0.6` in a separate PR, e.g. this will be removed in `v1`?

src/setfit/trainer.py
```diff
@@ -226,172 +206,126 @@ def call_model_init(self, params: Optional[Dict[str, Any]] = None):
...
 
         self._validate_column_mapping(self.train_dataset)
         train_dataset = self.train_dataset
         if self.column_mapping is not None:
             logger.info("Applying column mapping to training dataset")
             train_dataset = self._apply_column_mapping(self.train_dataset, self.column_mapping)
 
-        x_train = train_dataset["text"]
-        y_train = train_dataset["label"]
+        x_train: List[str] = train_dataset["text"]
```
lewtun, line 250: Nice!

src/setfit/trainer.py
```diff
@@ -530,3 +464,52 @@ def push_to_hub(
...
+        seed: int = 42,
+        column_mapping: Optional[Dict[str, str]] = None,
+        use_amp: bool = False,
+        warmup_proportion: float = 0.1,
+        distance_metric: Callable = BatchHardTripletLossDistanceFunction.cosine_distance,
+        margin: float = 0.25,
+        samples_per_label: int = 2,
+    ):
+        warnings.warn(
+            "`SetFitTrainer` has been deprecated. Please use `Trainer` instead.", DeprecationWarning, stacklevel=2
```
lewtun, line 491: Similar idea here to warn the user when it will be removed
```suggestion
            "`SetFitTrainer` has been deprecated and will be removed in v2 of SetFit. Please use `Trainer` instead.", DeprecationWarning, stacklevel=2
```

src/setfit/trainer_distillation.py
```diff
@@ -67,174 +67,137 @@ def __init__(
...
+            train_objectives=[(train_dataloader, train_loss)],
+            epochs=args.embedding_num_epochs,
+            optimizer_params={"lr": args.body_embedding_learning_rate},
+            warmup_steps=warmup_steps,
+            show_progress_bar=args.show_progress_bar,
+            use_amp=args.use_amp,
+        )
+
+
+class DistillationSetFitTrainer(DistillationTrainer):
```
lewtun, line 164: Should we add a deprecation warning for this too?

src/setfit/training_args.py
```diff
@@ -0,0 +1,93 @@
...
+import inspect
+from copy import copy
+from dataclasses import dataclass, field, fields
+from typing import Any, Callable, Dict, Tuple, Union
+
+from sentence_transformers.losses.BatchHardTripletLoss import BatchHardTripletLossDistanceFunction
+
+
+@dataclass
+class TrainingArguments:
```
lewtun, line 12: Very elegant implementation!

tests/test_deprecated_trainer.py
```diff
@@ -0,0 +1,412 @@
...
+from setfit.modeling import SetFitModel, SupConLoss
+from setfit.trainer import SetFitTrainer
+from setfit.utils import BestRun
+
+
+logging.set_verbosity_warning()
+logging.enable_propagation()
+
+
+class SetFitTrainerTest(TestCase):
```
lewtun, line 20: Note to our future selves - at some point we should refactor all the tests to follow the `unittest.TestCase` grouping instead of single functions. Personally, I think this makes it easier to understand what parts of the code are being tested :)

tests/test_training_args.py
```diff
@@ -0,0 +1,113 @@
...
+        with pytest.raises(ValueError):
+            TrainingArguments(warmup_proportion=-0.1)
+
+    def test_training_args_batch_sizes(self):
+        batch_size_A = 12
+        batch_size_B = 4
+        batch_size_C = 6
+
+        args = TrainingArguments(batch_size=batch_size_A)
+        assert args.batch_size == (batch_size_A, batch_size_A)
```
lewtun, line 24: If I'm not mistaken, using `self.assertEqual` is the idiomatic way to check these cases, e.g.

```suggestion
        self.assertEqual(args.batch_size, (batch_size_A, batch_size_A))
```

tests/test_training_args.py
```diff
@@ -0,0 +1,113 @@
+from unittest import TestCase
+
+import pytest
+
+from setfit.training_args import TrainingArguments
+
+
+class TestTrainingArguments(TestCase):
+    def test_training_args_raises_error_with_wrong_warmup_proportion(self):
```
lewtun, line 9: Since your class is called `TestTrainingArguments`, I think you can remove `training_args` from the function names
```suggestion
    def test_raises_error_with_wrong_warmup_proportion(self):
```

lewtun (approved):
Wow, this is a real tour de force of a PR @tomaarsen - great job refactoring all this complexity into clean APIs 🚀 !

Overall the code looks great and most of my comments are just nits + some suggestions for the deprecation warnings. I'll respond to your discussion points in a separate comment

lewtun:
Regarding the discussion points:

> How should users be notified on how to update their code? A section in the README? Via descriptive deprecation warnings? A link to this PR in the release notes?

In `transformers` we usually do this in two ways: add descriptive deprecations warnings that feature X will be removed in the next major version of the lib & add a "Breaking changes" section to the release notes with a link to this PR ([example](https://github.com/huggingface/transformers/releases/tag/v4.25.1)). I think we can adopt the same strategy here :)

> What should the deprecation timeline be? i.e. when do we remove SetFitTrainer altogether? We should include this in the deprecation warning messages.

I would like to remove it in v1 if possible, so one approach for that could be to make a v0.6 or v0.7 release that has all the deprecation warnings included. Alternatively, we could merge this in it's current form and tell users it will be removed in v2

> Should we simply deprecate Trainer.freeze() and Trainer.unfreeze? They should not be needed anymore, and they are simply equivalent to calling trainer.model.freeze() and trainer.model.unfreeze(). I'm in favour of the deprecation.

Yes, please deprecate!

> Are we a fan of the name end_to_end for training a differentiable model in full, i.e. not freezing the body when training the classifier?

I think it's OK, but happy to change it if you have a different name :)

> Should other arguments to Trainer be absorbed into TrainingArguments? In particular, loss_class?

Yes, I've been wondering this too - I think it would be good to put all non-dataset-dependent objects into `TrainingArguments` - WDYT? I think technically we could also put `metric` and `column_mapping` there too right?

tomaarsen:
Merged 9b7f74e56ab10253f7465e8176484fbaa36e1170 into this PR via aab23771fef1b86b8750497d8ceffef241b6004b.

Thanks for the detailed comments Lewis!
We can add [PendingDeprecationWarning](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) for the soon-to-be-removed features. 

Beyond that, I'm down to move `metric`, `loss` and `column_mapping` to `TrainingArguments`.

I'll try to get to work on your comments soon.

src/setfit/modeling.py
```diff
@@ -280,25 +281,29 @@ def fit(
         self,
         x_train: List[str],
         y_train: Union[List[int], List[List[int]]],
-        num_epochs: int,
-        batch_size: Optional[int] = None,
-        learning_rate: Optional[float] = None,
-        body_learning_rate: Optional[float] = None,
+        classifier_num_epochs: int,
+        classifier_batch_size: Optional[int] = None,
+        body_classifier_learning_rate: Optional[float] = None,
```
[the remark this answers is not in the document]
tomaarsen: That would make more sense. The reason I went for `body_classifier_learning_rate` is because it allows for [this](https://github.com/tomaarsen/setfit/blob/dee70b151e46932049739acf59dd9a80ec45a85e/src/setfit/trainer.py#L327).
This unpacking sends the variable from [here](https://github.com/tomaarsen/setfit/blob/dee70b151e46932049739acf59dd9a80ec45a85e/src/setfit/training_args.py#L27-L32) to the fit method.

I can go either way with this change, changing the variable requires for changing the way the `fit` is called (but maybe explicitly calling the `fit` with specific arguments clearly is more readable than just sending all training arguments into the `fit` method).

Alright, I talked myself into changing this, haha.

src/setfit/modeling.py
```diff
@@ -378,25 +390,40 @@ def freeze(self, component: Optional[Literal["body", "head"]] = None) -> None:
...
+        if (component is None or component == "head") and self.has_differentiable_head:
             self._freeze_or_not(self.model_head, to_freeze=True)
 
-    def unfreeze(self, component: Optional[Literal["body", "head"]] = None) -> None:
+    def unfreeze(
+        self, component: Optional[Literal["body", "head"]] = None, keep_body_frozen: Optional[bool] = None
+    ) -> None:
+        if keep_body_frozen is not None:
+            warnings.warn(
+                '`keep_body_frozen` is deprecated. Please either pass "head", "body" or no arguments to unfreeze both.',
```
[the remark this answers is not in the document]
tomaarsen: Done!

src/setfit/trainer.py
```diff
@@ -1,5 +1,13 @@
 import math
-from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
+import warnings
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
+
+
+# Google Colab runs on Python 3.7, so we need this to be compatible
```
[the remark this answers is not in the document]
tomaarsen: Let's shelve this for when we remove Python 3.7 support altogether in the coming months.

src/setfit/trainer_distillation.py
tomaarsen: We should! I missed this. Thanks

src/setfit/trainer.py
tomaarsen: We've discussed this privately, but for others reading: we'll go with a deprecation warning starting from v1 and complete removal at v2 (or at our discretion).

src/setfit/trainer.py
tomaarsen: Done!

src/setfit/training_args.py
tomaarsen: Thank you thank you

tests/test_training_args.py
tomaarsen: For `unittest` it indeed is. I'm used to the `pytest` approach which involves `assert` calls instead. I'll update this for consistency: better to go all `unittest` than mixing and matching too much.

tests/test_training_args.py
tomaarsen: Done!

tests/test_deprecated_trainer.py
tomaarsen: Noted! Beyond that, via `pytest` you can order specific classes to be tested, even if a file has multiple of these classes. Can definitely help speed up development if you don't have to wait for additional unrelated tests to pass.

tomaarsen:
`metric` and `column_mapping` are tricky to move to `TrainingArguments`. 

Let me elaborate why with an example.
```python
trainer = Trainer(
    model,
    args=args,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
)
trainer.train()
```
is equivalent to
```python
trainer = Trainer(
    model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
)
trainer.train(args)
```
in this PR.

When the `train` method is given some args, it will use those for that training call **only**.
However, if I move `metric` and/or `column_mapping`, then other methods (`evaluate`, `_validate_column_mapping`) will require a `TrainingArgument` instance.
This either means:
1. Requiring `evaluate` and `_validate_column_mapping` to be passed a `TrainingArgument` instance: highly unnecessary and would just be unintuitive.
2. Making `evaluate` and `_validate_column_mapping` use `self.args`, and using `self.args = TrainingArguments()` (i.e. default values) if `Trainer` is not passed args directly:
    I think this would quickly become very unintuitive if someone uses the second snippet from above, as the default from `self.args` will be used in `evaluate` and `_validate_column_mapping` rather than the more recently passed `args` in `trainer.train()`.
3. Require `args` to be passed to `TrainingArguments`: This is a decent solution, but it prevents the second snippet above from being possible, while that is the intuitive and expected approach for hyperparameter tuning. 
4. Making it so `args` passed to `trainer.train` updates the `self.args` permanently: I think this is unintuitive, and it prevents e.g. `trainer.evaluate()` before `trainer.train(args)` is called, as `self.args` will still be `None`. 

I can't think of any approach that doesn't have a glaring downside. Perhaps you have any ideas?

I'll move `loss` to `TrainingArguments` though, as that one is only used inside of the `train` method (and it's "children").

- Tom Aarsen

tomaarsen:
All of the TODO's are handled. @lewtun and I have decided that the updated README/notebooks/scripts will be implemented via a different PR.

There's one more thing that I'd like to get some feedback on, see the comment below.

src/setfit/modeling.py
```diff
@@ -376,41 +379,89 @@ def _prepare_optimizer(
...
+        """Set `requires_grad=not to_freeze` for all parameters in `model`"""
         for param in model.parameters():
             param.requires_grad = not to_freeze
 
-    def predict(self, x_test: List[str], as_numpy: bool = False) -> Union[torch.Tensor, "ndarray"]:
-        embeddings = self.model_body.encode(
-            x_test,
-            normalize_embeddings=self.normalize_embeddings,
-            convert_to_tensor=self.has_differentiable_head,
+    def encode(self, inputs: List[str]) -> Union[torch.Tensor, np.ndarray]:
```
tomaarsen, line 434: Although `SentenceTransformer` uses `encode`, so we may want to stick with that, we really are making an *embedding* moreso than an *encoding*. Perhaps this method should be called `embed`, e.g. `model.embed(["My input sentence"])`.

tomaarsen:
I've included a fix for #320 in this PR. Long story short, the DistillationTrainer now no longer requires a `label` column, and the `train_embeddings` and `train_classifier` methods for it no longer accept a `y_train` parameter. I've also removed the option to provide `losses.BatchAllTripletLoss`, `losses.BatchHardTripletLoss`, `losses.BatchSemiHardTripletLoss`, `losses.BatchHardSoftMarginTripletLoss` and `SupConLoss`, as I don't believe that the distillation training worked for these (it just took the input labels rather than learning using the teacher model).

tomaarsen:
Slight "issue" I've found: Providing a method like this:
```python
def hp_space(trial):  # Training parameters
    return {
        "body_learning_rate": trial.suggest_float("learning_rate", 1e-6, 1e-4, log=True),
        "num_epochs": trial.suggest_int("num_epochs", 1, 5),
        "batch_size": trial.suggest_categorical("batch_size", [4, 8, 16, 32, 64]),
        "seed": trial.suggest_int("seed", 1, 40),
        "num_iterations": trial.suggest_categorical("num_iterations", [5, 10, 20]),
        "max_iter": trial.suggest_int("max_iter", 50, 300),
        "solver": trial.suggest_categorical("solver", ["newton-cg", "lbfgs", "liblinear"]),
    }
```
will just quietly ignore the `body_learning_rate`, `num_epochs` and `batch_size` hyperparameters. They'll be set in the `TrainingArguments`, but internally only e.g. `body_embedding_learning_rate` and `classifier_num_epochs` are used. 
Beyond that, providing a parameter that isn't included in the `TrainingArguments` at all will not result in a warning or exception, e.g. if you include `learning_rate` or `lr`. This is because some hyperparameters need to be passed along to `model_init` instead.

Perhaps the correct approach is to always send a dictionary of parameters to `model_init`, and let the user create a `TrainingArguments` instance there? Currently, the hyperparameters are applied a bit hackishly like this:
https://github.com/huggingface/setfit/blob/c6f5782ccb42ab0ed3cd4a934dfa6f6db1de46b5/src/setfit/trainer.py#L160-L163

I don't think there's a really convenient way that is also backwards compatible. I'll look into it next week, and otherwise we just try to document it well, i.e. that users must use `body_embedding_learning_rate` in the `hp_space` method..

tomaarsen:
I'd like for #332 and #331 to be tackled first.

grofte:
Sorry for intruding on your PR. But if you are adding TrainingArguments, does this mean the user can change the optimizer to Adafactor as described here: https://huggingface.co/docs/transformers/perf_train_gpu_one#adafactor

tomaarsen:
Not as it is right now, but this would not be a very challenging addition.

tomaarsen:
New updates:
* Implement `TrainerControl`, `TrainerState`, and `CallbackHandler` from Transformers.
* Implement checkpointing
* Implement logging
* Add evaluation during training (with logging)
* Definitely works for `EarlyStoppingCallback`, `WandbCallback`.

PR #268, opened 2023-01-12
Add training option `unique_pairs`

danstan5:
Derived from #259

Changes:
* Updates the existing sentence pair generator: rather than random sampling an iterator returns new +/- combinations.
* Pairs are generated upto a `max_pairs` count (set by `num_iterations`). If all unique pairs have been generated and `unique_pairs=True` it will stop at this point (logging a warning).
* Pos/ neg pairs are generated separately to ensure positive pairs are not under-sampled (in case of high no. of classes). When `unique_pairs=True` neg/ pos pair counts will be remain balanced until all unique positive pairs have been added, then only unique neg pairs will continue to be added. Note: if concerned about an imbalance of negative pairs (in case of high no. of classes) `num_iterations` can simply be decreased to balance this as desired.
* Positive pair generator iterates through each class in turn extracting pairs. This ensures maximum variety of samples across different classes.

Overall these sampler changes would improve reproducibility + provides more representative sampling of the dataset, useful in cases of strong class imbalance or samples >> no. iterations training. 

`unique_pairs` adds an option for efficient embedding training by maximising only the available data for quicker training (testing for confirmation on this to follow..)

Still todo:
- [x] Add tests
- [x] Update for multi-label
- [x] Run performance evaluation on `test_set` - runtime/ accuracy comparisons

@tomaarsen keen to get your thoughts on this so far 👍

### DIFF
--- a/src/setfit/sampler.py
+++ b/src/setfit/sampler.py
@@ -1,9 +1,9 @@
-from typing import Generator, Iterable, List
-import numpy as np
-from torch.utils.data import  IterableDataset
 from itertools import zip_longest
+from typing import Generator, Iterable, List
 
+import numpy as np
 from sentence_transformers import InputExample
+from torch.utils.data import IterableDataset
 
 from . import logging
 
@@ -31,89 +31,6 @@ def shuffle_combinations(iterable: Iterable, replacement: bool = False) -> Gener
         yield iterable[_idx], iterable[idx]
 
 
-def positive_sentence_pairs_generate(
-    sentences: np.ndarray, labels: np.ndarray, max_pairs: int, unique_pairs: bool = False, multilabel: bool = False
-) -> List[InputExample]:
-    """Generates all unique or upto a max no. of combinations of positive sentence pairs.
-
-    Samples positive combinations of sentences (without replacement) and maximises
-        sampling of different classes in the pairs being generated.
-
-    Args:
-        sentences: an array of all sentences
-        labels: an array of the label_id for each item in `sentences`
-        max_pairs: returns when this many pairs are generated
-        unique_pairs: if true will return sentences if all unique combinations,
-            before max_pairs count is reached
-        multilabel: set to process "multilabel" labels array
-
-    Returns:
-        List of positive sentence pairs (upto the no. of unique_pairs or max_pairs)
-    """
-    pairs = []
-    if multilabel:
-        label_ids = np.arange(labels.shape[1])  # based on index = 1,0
-    else:
-        label_ids = np.unique(labels)  # based on class int
-    while True:
-        positive_combinators = []
-        for _label in label_ids:
-            if multilabel:
-                label_sentences = sentences[np.where(labels[:, _label] == 1)[0]]
-            else:
-                label_sentences = sentences[np.where(labels == _label)]
-            positive_combinators.append(shuffle_combinations(label_sentences, replacement=True))
-
-        for pos_pairs in zip_longest(*positive_combinators):
-            for pos_pair in pos_pairs:
-                if pos_pair is not None:
-                    pairs.append(InputExample(texts=[*pos_pair], label=1.0))
-                    if len(pairs) == max_pairs:
-                        return pairs
-        if unique_pairs:
-            break
-    logger.warning(f"** All ({len(pairs):,}) positive unique pairs generated")
-    return pairs
-
-
-def negative_sentence_pairs_generate(
-    sentences: np.ndarray,
-    labels: np.ndarray,
-    max_pairs: int,
-    unique_pairs: bool = False,
-    multilabel: bool = False,
-) -> List[InputExample]:
-    """Generates all or upto a max sample no. of negative combinations.
-
-    Randomly samples negative combinations of sentences (without replacement)
-
-    Args:
-        sentences: an array of all sentences
-        labels: an array of the label_id for each item in `sentences`
-        max_pairs: returns when this many pairs are generated
-        unique_pairs: if true will return sentences if all unique combinations,
-            before max_pairs count is reached
-        multilabel: set to process "multilabel" labels array
-
-    Returns:
-        List of negative sentence pairs (upto the no. of unique_pairs or max_pairs)
-    """
-    pairs = []
-    sentence_labels = list(zip(sentences, labels))
-    while True:
-        for (_sentence, _label), (sentence, label) in shuffle_combinations(sentence_labels):
-            # logical_and checks if labels are both set for each class
-            if (multilabel and not any(np.logical_and(_label, label))) or (not multilabel and _label != label):
-                pairs.append(InputExample(texts=[_sentence, sentence], label=0.0))
-                if len(pairs) == max_pairs:
-                    return pairs
-        if unique_pairs:
-            break
-    logger.warning(f"** All ({len(pairs):,}) negative unique pairs generated")
-    return pairs
-
-
-
 class ConstrastiveDataset(IterableDataset):
     def __init__(self, examples, num_iterations, unique_pairs, multilabel):
         """Generates positive and negative sentence pairs for contrastive learning.
@@ -123,38 +40,139 @@ class ConstrastiveDataset(IterableDataset):
             num_iterations: sets the number of contastive sample pairs to be generated
             unique_pairs: when true will only return upto the number of unique sentence
                 pair combinations avaliable
+            multilabel: set to process "multilabel" labels array
         """
         super().__init__()
 
-        max_pairs = num_iterations * len(examples)
-        sentences = np.array([s.texts[0] for s in examples])
-        labels = np.array([s.label for s in examples])
+        self.pos_index = 0
+        self.neg_index = 0
+        self.multilabel = multilabel
+        self.unique_pairs = unique_pairs
+        self.sentences = np.array([s.texts[0] for s in examples])
+        self.labels = np.array([s.label for s in examples])
+        self.max_pairs = num_iterations * len(examples)
 
-        positive_pairs = positive_sentence_pairs_generate(sentences, labels, max_pairs, unique_pairs, multilabel)
-        negative_pairs = negative_sentence_pairs_generate(sentences, labels, max_pairs, unique_pairs, multilabel)
+        # generate dataset so __len__ method can be used
+        self.generate_sentence_pairs()
 
-        if unique_pairs:
+    def generate_sentence_pairs(self) -> None:
+        """Generates a new batch of positive and negative sentence pairs.
+
+        Note: pos_index/ neg_index keep the position of pairs being generated.
+        """
+        positive_pairs = self.positive_sentence_pairs(self.max_pairs, self.unique_pairs)
+        negative_pairs = self.negative_sentence_pairs(self.max_pairs, self.unique_pairs)
+
+        if self.unique_pairs:
             extra_pairs = abs(len(positive_pairs) - len(negative_pairs))
 
             if len(positive_pairs) > len(negative_pairs):
-                logger.warning("** Oversampling negative pairs to balance contrastive training samples.")
-                negative_pairs += negative_sentence_pairs_generate(sentences, labels, extra_pairs, False, multilabel)
+                logger.warning(
+                    f"** Oversampling ({extra_pairs:,}) negative pairs to balance contrastive training samples."
+                )
+                negative_pairs += self.negative_sentence_pairs(max_pairs=extra_pairs, unique_pairs=False)
 
             if len(negative_pairs) > len(positive_pairs):
-                logger.warning("** Oversampling positive pairs to balance contrastive training samples.")
-                positive_pairs += positive_sentence_pairs_generate(sentences, labels, extra_pairs, False, multilabel)
+                logger.warning(
+                    f"** Oversampling ({extra_pairs:,}) positive pairs to balance contrastive training samples."
+                )
+                positive_pairs += self.positive_sentence_pairs(max_pairs=extra_pairs, unique_pairs=False)
 
         self._num_pairs = len(positive_pairs) + len(negative_pairs)
         self.positive_pairs = positive_pairs
         self.negative_pairs = negative_pairs
 
+    def positive_sentence_pairs(self, max_pairs: int, unique_pairs: bool = False) -> List[InputExample]:
+        """Generates all unique or upto a max no. of combinations of positive sentence pairs.
+
+        Samples positive combinations of sentences (without replacement) and maximises
+            sampling of different classes in the pairs being generated.
+
+        Args:
+            max_pairs: returns when this many pairs are generated
+            unique_pairs: if true will return sentences if all unique combinations,
+                before max_pairs count is reached
+
+        Returns:
+            List of positive sentence pairs (upto the no. of unique_pairs or max_pairs)
+        """
+        labels = self.labels
+        sentences = self.sentences
+        multilabel = self.multilabel
+        pairs = []
+
+        if multilabel:
+            label_ids = np.arange(labels.shape[1])  # based on index = 1,0
+        else:
+            label_ids = np.unique(labels)  # based on class int
+        while True:
+            index = 0
+            positive_combinators = []
+            for _label in label_ids:
+                if multilabel:
+                    label_sentences = sentences[np.where(labels[:, _label] == 1)[0]]
+                else:
+                    label_sentences = sentences[np.where(labels == _label)]
+                positive_combinators.append(shuffle_combinations(label_sentences, replacement=True))
+
+            for pos_pairs in zip_longest(*positive_combinators):
+                for pos_pair in pos_pairs:
+                    if pos_pair is not None:
+                        index += 1
+                        if index > self.pos_index:
+                            pairs.append(InputExample(texts=[*pos_pair], label=1.0))
+                            if len(pairs) == max_pairs:
+                                self.pos_index = index
+                                return pairs
+            self.pos_index = 0
+            if unique_pairs:
+                break
+        logger.warning(f"** All ({len(pairs):,}) positive unique pairs generated")
+        return pairs
+
+    def negative_sentence_pairs(self, max_pairs: int, unique_pairs: bool = False) -> List[InputExample]:
+        """Generates all or upto a max sample no. of negative combinations.
+
+        Randomly samples negative combinations of sentences (without replacement).
+
+        Args:
+            max_pairs: returns when this many pairs are generated
+            unique_pairs: if true will return sentences if all unique combinations,
+                before max_pairs count is reached
+
+        Returns:
+            List of negative sentence pairs (upto the no. of unique_pairs or max_pairs)
+        """
+        multilabel = self.multilabel
+        pairs = []
+
+        sentence_labels = list(zip(self.sentences, self.labels))
+        while True:
+            index = 0
+            for (_sentence, _label), (sentence, label) in shuffle_combinations(sentence_labels):
+                # logical_and checks if labels are both set for each class
+                if (multilabel and not any(np.logical_and(_label, label))) or (not multilabel and _label != label):
+                    index += 1
+                    if index > self.neg_index:
+                        pairs.append(InputExample(texts=[_sentence, sentence], label=0.0))
+                        if len(pairs) == max_pairs:
+                            self.neg_index = index
+                            return pairs
+            self.neg_index = 0
+            if unique_pairs:
+                break
+        logger.warning(f"** All ({len(pairs):,}) negative unique pairs generated")
+        return pairs
 
     def __iter__(self):
-        # generates one of each in turn
         for pos_pair, neg_pair in zip(self.positive_pairs, self.negative_pairs):
+            # generates one of each in turn
             yield pos_pair
             yield neg_pair
 
+        if self.pos_index or self.neg_index:
+            # not all pairs combinations sampled so continues from last index
+            self.generate_sentence_pairs()
 
     def __len__(self):
         return self._num_pairs
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -82,7 +82,7 @@ class SetFitTrainer:
         eval_dataset: Optional["Dataset"] = None,
         model_init: Optional[Callable[[], "SetFitModel"]] = None,
         metric: Union[str, Callable[["Dataset", "Dataset"], Dict[str, float]]] = "accuracy",
-        loss_class: Optional["nn.Module"] = None,
+        loss_class: Optional[Any] = None,
         num_iterations: int = 20,
         num_epochs: int = 1,
         learning_rate: float = 2e-5,
@@ -314,6 +314,8 @@ class SetFitTrainer:
             logger.warning("No `loss_class` detected! Using `CosineSimilarityLoss` as the default.")
             self.loss_class = losses.CosineSimilarityLoss
 
+        multilabel = True if self.model.multi_target_strategy is not None else False
+
         num_epochs = num_epochs or self.num_epochs
         batch_size = batch_size or self.batch_size
         learning_rate = learning_rate or self.learning_rate
@@ -333,19 +335,17 @@ class SetFitTrainer:
                 SupConLoss,
             ]:
                 train_data_sampler = SentenceLabelDataset(train_examples, samples_per_label=self.samples_per_label)
-                train_dataloader = DataLoader(
-                    train_data_sampler,
-                    batch_size=min(batch_size, len(train_data_sampler)),
-                    drop_last=True
-                )
+                batch_size = min(batch_size, len(train_data_sampler))
+                train_dataloader = DataLoader(train_data_sampler, batch_size=batch_size, drop_last=True)
             else:  # setfit default constrastive pairs generator
-                multilabel = True if self.model.multi_target_strategy is not None else False
-                train_data_sampler = ConstrastiveDataset(train_examples, self.num_iterations, self.unique_pairs, multilabel)
+                train_data_sampler = ConstrastiveDataset(
+                    train_examples, self.num_iterations, self.unique_pairs, multilabel
+                )
                 train_dataloader = DataLoader(train_data_sampler, batch_size=batch_size, drop_last=False)
 
             total_train_steps = len(train_dataloader) * num_epochs
             logger.info("***** Running training *****")
-            logger.info(f"  Num examples = {len(train_data_sampler)}")
+            logger.info(f"  Num examples per epoch = {len(train_data_sampler)}")
             logger.info(f"  Num epochs = {num_epochs}")
             logger.info(f"  Total optimization steps = {total_train_steps}")
             logger.info(f"  Total train batch size = {batch_size}")


src/setfit/modeling.py
```diff
@@ -633,30 +634,87 @@ def forward(self, sentence_features, labels=None, mask=None):
         return loss
 
 
-def sentence_pairs_generation(sentences, labels, pairs):
-    # Initialize two empty lists to hold the (sentence, sentence) pairs and
-    # labels to indicate if a pair is positive or negative
+def positive_sentence_pairs_generate(
+    sentences: np.ndarray[str], labels: np.ndarray[int], max_pairs: int, unique_pairs: bool
```
tomaarsen, line 638:
```suggestion
    sentences: "np.ndarray[str]", labels: "np.ndarray[int]", max_pairs: int, unique_pairs: bool
```
Alternatively, `ArrayLike` may be applicable? https://numpy.org/doc/stable/reference/typing.html#numpy.typing.ArrayLike

Otherwise we'll run into errors (as you can see with the tests)

src/setfit/modeling.py
```diff
@@ -633,30 +634,87 @@ def forward(self, sentence_features, labels=None, mask=None):
...
+                        return pairs
+
+        if unique_pairs:
+            break
+    logger.warning(f"** All ({len(pairs):,}) positive unique pairs generated")
+    return pairs
+
+
+def negative_sentence_pairs_generate(
+    sentences: np.ndarray[str], labels: np.ndarray[int], max_pairs: int, unique_pairs: bool
```
tomaarsen, line 676:
```suggestion
    sentences: "np.ndarray[str]", labels: "np.ndarray[int]", max_pairs: int, unique_pairs: bool
```
idem.

src/setfit/modeling.py
```diff
@@ -633,30 +634,87 @@ def forward(self, sentence_features, labels=None, mask=None):
...
+                if len(pairs) == max_pairs:
+                    return pairs
+        if unique_pairs:
+            break
+    logger.warning(f"** All ({len(pairs):,}) negative unique pairs generated")
     return pairs
 
 
+def sentence_pairs_generation(
+    sentences: np.ndarray[str], labels: np.ndarray[int], max_pairs: int, unique_pairs: bool
```
tomaarsen, line 707:
```suggestion
    sentences: "np.ndarray[str]", labels: "np.ndarray[int]", max_pairs: int, unique_pairs: bool
```
idem.

tomaarsen (changes requested):
I'll have to look at it more closely some other time, but the writeup looks good! I'm glad my concerns are noted.

I do have some small nitpicks already, see my other comments. These are causing test failures.

src/setfit/modeling.py
danstan5: Ahh yes. I'm happy with those changes or just back to `np.ndarray` only and will add the array-types in the docstring? 
The `npt.ArrayLike` loads stuff pretty ugly and I've not seen in other projects before...

tomaarsen: I also have never used `npt.ArrayLike` before. I think `"np.ndarray[int]"` is more than fine for now.

src/setfit/modeling.py
```diff
@@ -633,30 +634,87 @@ def forward(self, sentence_features, labels=None, mask=None):
...
+                if len(pairs) == max_pairs:
+                    return pairs
+        if unique_pairs:
+            break
+    logger.warning(f"** All ({len(pairs):,}) negative unique pairs generated")
     return pairs
 
 
+def sentence_pairs_generation(
+    sentences: "np.ndarray[str]", labels: "np.ndarray[int]", max_pairs: int, unique_pairs: bool
```
tomaarsen, line 707:
```suggestion
    sentences: "np.ndarray[str]", labels: "np.ndarray[int]", max_pairs: int, unique_pairs: bool = False
```
If we set a default here, then people's code shouldn't break if they use `sentence_pairs_generation` directly for some reason. This is also causing some test failures.

danstan5: Because the logic of `sentence_pairs_generation` has changed slightly will need to update the test for `test_sentence_pairs_generation`, will do this when I update/ create new tests.

danstan5:
* I've rolled **multi-label** into the `sentence_pairs_generation` function. I think this is cleaner as can now see the logic to generate pairs is exactly the same in both cases. The only difference is just for formatting `labels` → pos/ neg pairs.
* Original tests have been fixed so passing again + some new ones added for `unique_pairs`.

tomaarsen:
First of all, these changes look great! I way prefer simply having 
```python
train_examples = sentence_pairs_generation(...)
```
rather than
```python
train_examples = []
for _ in range(...):
    train_examples = sentence_pairs_generation(..., train_examples)
```

I ran some quick experiments locally, and I noticed an interesting quirk. Before I stress you out: this quirk also exists on main. If I increase `num_epochs` from 1 to 2, then it runs two epochs instead of one like expected, but it *also* increments the number of iterations per epoch?

So, whereas I might get 34 training steps times 1 epoch (so, 34 steps total) with `num_epochs=1`, I'll get 68 training steps times 2 epochs for `num_epochs=2` (so, 136 steps total). 
<details><summary>Simple script to reproduce the quirk</summary>

```python

)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    column_mapping={"sentence": "text", "label": "label"},  # Map dataset columns to text/label expected by trainer
    num_epochs=1, # <- Change 1 to see the quirk
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()
```
</details>

It's caused by this line:
https://github.com/huggingface/setfit/blob/f777c2c60b270604dae0dc1db4eea815e8c9019d/src/setfit/trainer.py#L378
The `train_steps` is defined like so:
https://github.com/huggingface/setfit/blob/f777c2c60b270604dae0dc1db4eea815e8c9019d/src/setfit/trainer.py#L366
Which explains the behaviour. In my opinion, the number of steps per epoch should be equivalent to the number of datapoints in the dataloader. That is exactly the behaviour if `steps_per_epoch` is not supplied. So, could you remove `steps_per_epoch`?

## Experiments

As mentioned, I ran some experiments. In particular, I used `unique_pairs=True` and `unique_pairs=False` while keeping the number of total optimization steps roughly equivalent between the two scripts. I also applied the fix of the quirk mentioned above. 
<details><summary>unique_pairs = False w. 3 epochs of 640 steps (40 batches) each (1920 total optimization steps) (Baseline)</summary>

```python

)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    column_mapping={"sentence": "text", "label": "label"},  # Map dataset columns to text/label expected by trainer
    num_epochs=3,
    unique_pairs=False,
    seed=1,
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()
```
</details>

<details><summary>unique_pairs = True w. 14 epochs of 136 steps (9 batches) each (1904 total optimization steps)</summary>

```python

)

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    column_mapping={"sentence": "text", "label": "label"},  # Map dataset columns to text/label expected by trainer
    num_epochs=14,
    unique_pairs=True,
    seed=1,
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()
```
</details>

## Results

| Experiment                       | Evaluation Accuracy |
|----------------------------------|----------------------|
| unique_pairs = False w. 3 epochs (Baseline) | 86.697% (1.593)      |
| unique_pairs = True w. 14 epochs (This PR) | 87.461% (0.593)      |

Notes:
1. Evaluation accuracy is displayed as mean and standard deviation between 12 executions.
2. The seed on the `SetFitTrainer` is updated for every execution, but the seed on the `sample_dataset` calls stays the same. This way, all executions get the same input data.
3. Each of the experiments took roughly 1 minute to train, give or take 3 seconds.

The results are almost statistically significant at a p-value of .074996 (assuming p < .05 for statistical significance), but certainly very promising. Especially the difference in standard deviation is very promising for this PR, as it displays the expected behaviour that non-random sampling produces more stable and consistent results. I'm quite excited about it.

I'm considering running more experiments. Feel free to run your own ones, too!

- Tom Aarsen

tomaarsen:
## Experiments on emotion

I ran some more experiments. In particular, I used `unique_pairs=True` and `unique_pairs=False` while keeping the number of total optimization exactly equivalent between the two scripts. I also applied the fix of the quirk mentioned in my previous comment. 

These experiments differ from the previous experiments on sst2 in that each epoch ran the same number of examples, and that I used the same number of epochs between the experiments. Furthermore, sst2 is binary, while the emotion dataset contains 6 keys.
<details><summary>unique_pairs = False w. num_iterations=10 and 2 epochs of 960 steps (60 batches) each (1920 total optimization steps) (Baseline)</summary>

```python

dataset = load_dataset("SetFit/emotion")

for seed in range(12):
    )

    # Create trainer
    trainer = SetFitTrainer(
        model=model,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        num_epochs=2,
        num_iterations=10,
        unique_pairs=False,
        seed=seed
    )

    # Train and evaluate
    trainer.train()
    metrics = trainer.evaluate()

    print(metrics)
```
</details>

<details><summary>unique_pairs = True w. num_iterations=10 and 2 epochs of 960 steps (60 batches) each (1920 total optimization steps)</summary>

```python

dataset = load_dataset("SetFit/emotion")

for seed in range(12):
    )

    # Create trainer
    trainer = SetFitTrainer(
        model=model,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        num_epochs=2,
        num_iterations=10,
        unique_pairs=True,
        seed=seed
    )

    # Train and evaluate
    trainer.train()
    metrics = trainer.evaluate()

    print(metrics)
```
</details>

## Results

| Experiment                       | Evaluation Accuracy |
|----------------------------------|----------------------|
| main branch w. num_iterations=10 and 2 epochs (Baseline) | 44.850% (1.648)      |
| unique_pairs = False w. num_iterations=10 and 2 epochs (Baseline) | 42.946% (1.615)      |
| unique_pairs = True w. num_iterations=10 and 2 epochs (This PR) | 48.717% (1.261)      |

Notes:
1. Evaluation accuracy is displayed as mean and standard deviation between 12 executions.
2. The seed on the `SetFitTrainer` is updated for every execution, but the seed on the `sample_dataset` calls stays the same. This way, all executions get the same input data.
3. Each of the experiments took roughly 1:45 minutes to train, give or take 3 seconds.
4. I included the first Baseline test to see whether `unique_pairs=False` would be equivalent to main.

image

These results are very much statistically significant! These seem like serious improvements. And again, we can note an increase in stability as `unique_pairs` is set to True through the decreased standard deviation.

However, from these results it is also clear that the performance of `unique_pairs=False` seems degraded relative to the main branch. The two sets of results are statistically significantly different, which is somewhat worrisome. I was expecting the behaviour of the main branch and `unique_pairs=False` to be equivalent. Could you elaborate on the differences?

I'm very positive about these results. 

- Tom Aarsen

tomaarsen:
Dan and I had a discussion on this topic, which I will share for others reading through this interesting PR:

We theorize that the reduction in accuracy when `unique_pairs=False` is due to how the new pairs are generated in a natural order e.g. `(0,0), (0,1), (0,2), (0,3), (0,4), (1,1), ...`, causing to non-representative sampling. This could be fixed by randomising the order of the combinations generated, but we'll have to figure out how to do this in a memory-efficient generator approach.

Additionally, Dan tested equal representation of positive pairs versus negative pairs compared to the `unique_pairs` approach from this PR. The `unique_pairs` approach often results in much more negative pairs due to how there are usually less possible positive pairs. He discovered that the 50:50 approach is generally much better than when the training samples are skewed towards negative. 
One possible solution here is to adapt the `unique_pairs` approach to compute `num_sample = min(max_num_positive_pairs, max_num_negative_pairs, user_specified_max_samples)`, and sample that many positive and that many negative pairs. 

Furthermore, I'm very interested in "freshly" generating samples per epoch. For example, if there are 20.000 possible negative pairs, and we sample 2.000 in a 20-epoch training scenario, then currently we use the same 2,000 samples for each of the 20 epochs. We might get an increase in performance if we can re-sample 2.000 from the 20.000 for every epoch again.

- Tom Aarsen

danstan5:
**Non-representative sampling fix** - `shuffle_combinations` function is used instead of `itertools.combinations` to shuffle generated pairs. This is based on numpy generated indicies & indexing and [looks to be the fastest method](https://stackoverflow.com/a/63694661/13730496) (I've tested generating upto ~10^8 combinations with no issues).

Overall the new sampler improves reproduciblity and samples labelled datasets more evenly, with slight accuracy & reducing std dev as a result:

| Emotion dataset (8 samples per class)  | Main Branch | New sampler    |
|-----------------------------------------------------------|-------------|----------------|
| 7 iterations (672 pairs, 42 optimization steps)     | 44.9 (4.3)  | **45.8 (3.1)** |
| 13 iterations (1,248 pairs, 78 optimization steps)  | 47.4 (4.8)  | **47.9 (3.7)** |
| 26 iterations (2,496 pairs, 156 optimization steps) | 47.9 (5.3)  | **50.1 (4.9)** |

**Setting up unique_pairs** - forcing use of equal pos/ neg samples shows better results but requires sampling "unique pairs" differently...
* 1 option is generate all positive pairs then same no. of negative. This (I will call the *min_unique_pairs* case for now) can result in small no. of training steps with v fast runtimes but sub-optimum accuracy
* Another alternative here would set to *max_unique_pairs* - get all negative pairs then just oversample positive pairs to balance. This uses the avaliable data better, but isn't strictly "unique pairs" anymore

Results on the multi-class (imbalanced) test datasets (opt_steps are a proxy for runtime), default 1 epoch:

| 8 samples per class | `emotion` opt_steps | `emotion` accuracy | `sst5` opt_steps | `sst5` accuracy | `ag_news` opt_steps | `ag_news` accuracy |
|---------------------------------------------|-----------------------|--------------------|--------------------|-----------------|-----------------------|--------------------|
| Unbalanced unique_pairs (all)               | 74                    | 46.4 (3.9)         | 51                 | 40.2 (2.8)      | 33                    | 82.2 (3.1)         |
| min_unique_pairs (undersample)             | 27                    | 42.4 (3.9)         | 23                 | 40.7 (2.5)      | 18                    | 81.2 (2.8)         |
| **3x epoch** min_unique_pairs (undersample)  | 81                    | 47.7 (4.7)         | 69                 | 43.4 (2.6)      | 54                    | 83.0 (3.1)         |
| max_unique_pairs (oversample)              | 120                   | **50.9 (3.7)**     | 80                 | **43.9 (2.4)**  | 48                    | **83.5 (2.8)**     |

Results show increasing no. of epochs improves accuracy in roughly similar runtime, but not as effective as more sampling. As noted in the last comment in this PR, generating samples for each epoch run would be nice addition, however epochs are handled by the `Model` class and torch `dataloader`, probably beyond the scope of this PR.
Within the current constraints I would recommend generating higher num iterations (for more samples) instead of increased num epochs to benefit from all of the labelled data available.

danstan5:
* `shuffle sample pair combinations`(https://github.com/huggingface/setfit/pull/268/commits/ca18e69a6bd3679c47e70bf24c5ce1a122d43e4c) shows *min unique pairs* logic
* `update unique_pairs logic`(https://github.com/huggingface/setfit/pull/268/commits/ced309c624092336982bdb8bcc38faaba0681b72) shows the *max unique pairs* logic

I'm lending towards the latter due to the higher accuracy and provides a natural cut-off to the no. of samples used in training. `num_iterations` can be reducing too speed up training if needed (and now with better sampling). @tomaarsen keen to get your thoughts on all the above

tomaarsen:
@danstan5 
My apologies, I lost sight of this PR. Let me dive back in.

My first hunch is to prefer oversampling rather than undersampling, for two reasons:
1. In few-shot it is crucial to take advantage of as much of our data as possible.
2. We can set an upper bound on oversampling using a user-provided sample maximum, but we can't as easily set a lower bound on undersampling.

I also want to point out that I intend for the next release to be `v1.0.0`, which means that I'm open to more breaking changes if we believe that they are beneficial for users.

- Tom Aarsen

danstan5:
Nice one. Strong agree on the oversampling.. struggling for good names though (`set_max_pairs` or something?)

Sure, I think it's best placed in a more major release as it will ultimately change existing results if your re-train setfit models. 
Also v1 is looking great so far! I will update this PR as needed after the refactor goes in. Let me know if have any further suggestions on it till then.

Thanks, Dan

tomaarsen:
Hello!

I've ran some more (much more thorough) experiments on this PR, as myself and others suspected that this PR might help with an ubiquitous overfitting problem on SetFit models. After all, better data gives better models, and conceptually this PR should result in better data.

## Experiment details
I have ran three sets of experiments:
* **A**: From the main branch at 78851287535305ef32f789c7a87004628172b5b6:
  ```cmd
  python .\scripts\setfit\run_fewshot.py --is_dev_set=True --sample_sizes 2 4 8 16 --batch_size 32
  ```
* **B**: From this PR at 3ce733f3ee5c96386a259fdd28c32d95add01e62, using `unique_pairs=False`:
  ```cmd
  python .\scripts\setfit\run_fewshot.py --is_dev_set=True --sample_sizes 2 4 8 16 --batch_size 32 --unique_pairs=False
  ```
* **C**: From this PR at 3ce733f3ee5c96386a259fdd28c32d95add01e62, using `unique_pairs=True`, i.e. the oversampling approach that you described:
  ```cmd
  python .\scripts\setfit\run_fewshot.py --is_dev_set=True --sample_sizes 2 4 8 16 --batch_size 32 --unique_pairs=True
  ```
  With the following changes in place:
  <details><summary>Click to see the diff</summary>

  ```diff
  diff --git a/src/setfit/modeling.py b/src/setfit/modeling.py
  index 0978061..d5f96a4 100644
  --- a/src/setfit/modeling.py
  +++ b/src/setfit/modeling.py
  @@ -771,6 +771,24 @@ def negative_sentence_pairs_generate(
       return pairs
   
   
  +def bounded_cycle(iterable, N):
  +    # bounded_cycle('ABCD', 10) --> A B C D A B C D A B
  +    n_yielded = 0
  +    saved = []
  +    for element in iterable:
  +        yield element
  +        n_yielded += 1
  +        if n_yielded >= N:
  +            return None
  +        saved.append(element)
  +    while saved:
  +        for element in saved:
  +            yield element
  +            n_yielded += 1
  +            if n_yielded >= N:
  +                return None
  +
  +
   def sentence_pairs_generation(
       sentences: np.ndarray,
       labels: np.ndarray,
  @@ -803,7 +821,7 @@ def sentence_pairs_generation(
               logger.warning("** Oversampling positive pairs to balance contrastive training samples.")
               positive_pairs += positive_sentence_pairs_generate(sentences, labels, extra_pairs, False, multilabel)
   
  -    return positive_pairs + negative_pairs
  +    return list(bounded_cycle(positive_pairs + negative_pairs, max_pairs * 2))
   
   
   def sentence_pairs_generation_cos_sim(sentences, pairs, cos_sim_matrix):
  ```
  </details>

  These changes ensure that all of the experiments run with **exactly** the same amount of training samples, even if less pairs would normally be selected when `unique_pairs` is set to `True` (which would normally be compensated with a higher `num_epochs`).

### Goals
I have also written a script to plot the results from the three separate `results` folders that were created from these three experiments. Ideally, we would like to see a situation where approach B is equivalent to approach A, i.e. if `unique_pairs=False`, then the existing performance of the main branch is preserved, **and** we would like to see approach C outperform the other two approaches.

### Results
<details><summary>Click here to see the resulting plots</summary>

bbc-news
enron_spam
imdb
sst2
student-question-categories
subj
toxic_conversations
TREC-QC

</details>

### Discussion
I am unable to see significant differences between the three approaches tested here, which is quite a shame. These results are quite surprising to me, as in my previous experiments [[1]](https://github.com/huggingface/setfit/pull/268#issuecomment-1382924584) [[2]](https://github.com/huggingface/setfit/pull/268#issuecomment-1382967119), I encountered improved performance when considering the same number of iterations.

I see three reasonable explanations:
1. I have made a mistake either in this experiment or in both of my prior ones.
2. Between acb27e8700642d32b7e742d97a82e76bccbf43fc...3ce733f3ee5c96386a259fdd28c32d95add01e62 in this PR, the performance gain was somehow lost.
3. Between the when I made my prior experiments, the main branch has been updated such that the baseline performance of the main branch has improved as well.

I'd like to further investigate option 2 this afternoon.

I'd like to notify @danielkorat and @MosheWasserb of these results, as I think they may benefit from the results of this experiment. I also think that @lewtun may be interested in this PR and its various experiments.

- Tom Aarsen

danstan5:
Hey @tomaarsen love these plots, can you share the script for them??

Your right there's nothing to statistically difference in these results, but this is because the default `num_iterations` is 20, the datasets are being oversampled anyway (so random sampling will look more suitably uniform). For example just checked on SST2 (8 samples) and `unique_pairs` -> 144 while `num_iterations=20` -> 640. 

The benefits of this PR only really only come at low iteration no. and the fact unique_pairs provides a nice natural limit if not sure how to set num_iterations. This is still useful for very large datasets or in hyperparameter test/ active-training loops etc.

tomaarsen:
To add on to my previous comment, it does not seem like the performance degraded since earlier in this PR. 

## Experiment
I ran another experiment:
* **D**: From this PR at 3ce733f3ee5c96386a259fdd28c32d95add01e62, using `unique_pairs=True`:
  ```
  python .\scripts\setfit\run_fewshot.py --is_dev_set=True --sample_sizes 2 4 8 16 --batch_size 32 --unique_pairs=True
  ```
  With the following changes in place:
  <details><summary>Click to see the diff</summary>
  
  ```diff
  diff --git a/src/setfit/modeling.py b/src/setfit/modeling.py
  index 9e8a04d..7af18ee 100644
  --- a/src/setfit/modeling.py
  +++ b/src/setfit/modeling.py
  @@ -717,6 +717,24 @@ def negative_sentence_pairs_generate(
       return pairs
   
   
  +def bounded_cycle(iterable, N):
  +    # bounded_cycle('ABCD', 10) --> A B C D A B C D A B
  +    n_yielded = 0
  +    saved = []
  +    for element in iterable:
  +        yield element
  +        n_yielded += 1
  +        if n_yielded >= N:
  +            return None
  +        saved.append(element)
  +    while saved:
  +        for element in saved:
  +            yield element
  +            n_yielded += 1
  +            if n_yielded >= N:
  +                return None
  +
  +
   def sentence_pairs_generation(
       sentences: np.ndarray,
       labels: np.ndarray,
  @@ -742,7 +760,7 @@ def sentence_pairs_generation(
       max_neg_pairs = (num_iterations * len(sentences) * 2) - len(positive_pairs)
       negative_pairs = negative_sentence_pairs_generate(sentences, labels, max_neg_pairs, unique_pairs, multilabel)
   
  -    return positive_pairs + negative_pairs
  +    return list(bounded_cycle(positive_pairs + negative_pairs, num_iterations * len(sentences) * 2))
   
   
   def sentence_pairs_generation_cos_sim(sentences, pairs, cos_sim_matrix):
  ```
  </details>

  As with **C**, these changes ensure the same number of iterations are used.

## Results
<details><summary>Click to see the results of the experiment</summary>

bbc-news
enron_spam
sst2
subj

</details>

## Discussion
As before, the performance seems nearly identical to the other results, despite now using the older version of this PR. 

I'm curious to hear your thoughts,

- Tom Aarsen

tomaarsen:
> So useful for very large datasets or in hyperparameter test/ active-training loops etc.

Wonderful! I'll try to see if I can experiment a bit further in those cases. I was under the impression that this PR was most interesting in small dataset situations. Most of all I'm a bit saddened that I can't seem to reproduce my earlier results where this PR seemed to result in a ~3% accuracy jump from 45% to 48% on the emotion dataset.

As for the script I used, I'll attach it here now, but I'll also fix up the comments, imports etc. (e.g. some comments talk about `create_summary_table.py`). I'll try to push it to the main branch as well, soon, as I've found it fairly useful.
<details><summary>Click to see plotting script</summary>

```python
import argparse
import json
import os
from pathlib import Path
from collections import defaultdict
from glob import glob
from typing import List, Tuple

import matplotlib.pyplot as plt
import pandas as pd
import string


def get_sample_sizes(path: str) -> List[str]:
    return sorted(list({int(name.split("-")[-2]) for name in glob(f"{path}/*/train-*-0")}))


def get_formatted_ds_metrics(path: str, dataset: str, sample_sizes: List[str]) -> Tuple[str, List[str]]:
    split_metrics = defaultdict(list)

    for sample_size in sample_sizes:
        result_jsons = sorted(glob(os.path.join(path, dataset, f"train-{sample_size}-*", "results.json")))
        for result_json in result_jsons:
            with open(result_json) as f:
                result_dict = json.load(f)

            metric_name = result_dict.get("measure", "N/A")
            split_metrics[sample_size].append(result_dict["score"])

    return metric_name, split_metrics


def plot_summary_comparison(paths: List[str]) -> None:
    dataset_to_df = defaultdict(pd.DataFrame)
    dataset_to_metric = {}
    for path_index, path in enumerate(paths):
        ds_to_metric, this_dataset_to_df = get_summary_df(path)
        for dataset, df in this_dataset_to_df.items():
            df["path_index"] = path_index
            dataset_to_df[dataset] = pd.concat((dataset_to_df[dataset], df))
        dataset_to_metric = dataset_to_metric | ds_to_metric

    # Prepare folder for storing figures
    image_dir = Path("scripts") / "images"
    image_dir.mkdir(exist_ok=True)
    new_version = max([int(path.name[2:]) for path in image_dir.glob("v_*/") if path.name[2:].isdigit()], default=0) + 1
    output_dir = image_dir / f"v_{new_version}"
    output_dir.mkdir()

    for dataset, df in dataset_to_df.items():
        columns = [column for column in df.columns if not column.startswith("path")]
        fig, axes = plt.subplots(ncols=len(columns), sharey=True)
        for column_index, column in enumerate(columns):
            ax = axes[column_index]

            # Set the y label only for the first column
            if column_index == 0:
                ax.set_ylabel(dataset_to_metric[dataset])

            # Set positions to 0, 0.25, ..., one position per boxplot
            # This places the boxplots closer together
            n_boxplots = len(df["path_index"].unique())
            allotted_box_width = 0.2
            positions = [allotted_box_width * i for i in range(n_boxplots)]
            ax.set_xlim(-allotted_box_width * 0.75, allotted_box_width * (n_boxplots - 0.25))

            # ax.set_xticks(range(n_boxplots), rotation="vertical")
            df[[column, "path_index"]].groupby("path_index", sort=True).boxplot(
                subplots=False, ax=ax, column=column, positions=positions
            )

            k_shot = column.split("-")[-1]
            ax.set_xlabel(f"{k_shot}-shot")
            if n_boxplots > 1:
                # If there are multiple boxplots, override the labels at the bottom generated by pandas
                if n_boxplots <= 26:
                    ax.set_xticklabels(string.ascii_uppercase[:n_boxplots])
                else:
                    ax.set_xticklabels(range(n_boxplots))
            else:
                # Otherwise, just remove the xticks
                ax.tick_params(labelbottom=False)

        if n_boxplots > 1:
            fig.suptitle(f"Comparison between various baselines on the {dataset}\ndataset under various $K$-shot conditions")
        else:
            fig.suptitle(f"Results on the {dataset} dataset under various $K$-shot conditions")
        fig.tight_layout()
        plt.savefig(str(output_dir / dataset))


def get_summary_df(path: str) -> None:
    """Given per-split results, creates a summary table of all datasets,
    with average metrics and standard deviations.

    Args:
        path: path to per-split results: either `scripts/{method_name}/{results}/{model_name}`,
            or `final_results/{method_name}/{model_name}.tar.gz`
    """

    sample_sizes = get_sample_sizes(path)
    header_row = ["dataset", "measure"]
    for sample_size in sample_sizes:
        header_row.append(f"{sample_size}_avg")
        header_row.append(f"{sample_size}_std")

    dataset_to_metric = {}
    dataset_to_df = {}
    for dataset in next(os.walk(path))[1]:
        metric_name, split_metrics = get_formatted_ds_metrics(path, dataset, sample_sizes)
        dataset_df = pd.DataFrame(split_metrics.values(), index=[f"{dataset}-{key}" for key in split_metrics]).T
        dataset_to_metric[dataset] = metric_name
        dataset_to_df[dataset] = dataset_df
    return dataset_to_metric, dataset_to_df


def main() -> None:
    parser = argparse.ArgumentParser()

    parser.add_argument("--paths", nargs="+", type=str)
    args = parser.parse_args()

    plot_summary_comparison(args.paths)


if __name__ == "__main__":
    main()
```

</details>

- Tom Aarsen

danstan5:
> Furthermore, I'm very interested in "freshly" generating samples per epoch

[These changes](https://github.com/huggingface/setfit/pull/268/files/3ce733f3ee5c96386a259fdd28c32d95add01e62..47b20cb2364498e6fa72f800ce267939156a6ac3) are a solution to the above:

- Moves contrastive pairs sampler logic from `.modelling` into a `.sampler` module
- Use a `ConstrastiveDataset` class that inherits from torch `IterableDataset` instead of functions. This matches exactly how sentence-transformers create training samples, see here calling the [non setfit sampler](https://github.com/huggingface/setfit/pull/268/files/3ce733f3ee5c96386a259fdd28c32d95add01e62..47b20cb2364498e6fa72f800ce267939156a6ac3#diff-8a5c0c6bad65c207df366694f985e2d8202bf4016694b02f17ce1c08ea595a48R337) uses same logic as a [setfit sampler](https://github.com/huggingface/setfit/pull/268/files/3ce733f3ee5c96386a259fdd28c32d95add01e62..47b20cb2364498e6fa72f800ce267939156a6ac3#diff-8a5c0c6bad65c207df366694f985e2d8202bf4016694b02f17ce1c08ea595a48R341).
- https://github.com/huggingface/setfit/pull/268/commits/47b20cb2364498e6fa72f800ce267939156a6ac3 Having this in a class, the state of generated samples picked up again between epochs, so it continues generating new samples between epochs if needed. 

Breaking changes:
- the `Dataloader` class [cannot shuffle](https://github.com/huggingface/setfit/pull/268/files/3ce733f3ee5c96386a259fdd28c32d95add01e62..47b20cb2364498e6fa72f800ce267939156a6ac3#diff-8a5c0c6bad65c207df366694f985e2d8202bf4016694b02f17ce1c08ea595a48L357-R344) when reading from the `iterableDataset` and now just draws [pos/neg pairs into the batches](https://github.com/huggingface/setfit/pull/268/files/3ce733f3ee5c96386a259fdd28c32d95add01e62..47b20cb2364498e6fa72f800ce267939156a6ac3#diff-5f33f86a0c6233d04f3fd59590720614adbd25f4d0beb5f8dcaf8c5fe81ca9b7R168-R171). *Need to review this impact on performance still.*

Impact:
- Expect for the shuffling ^(yet to be reviewed), this is no different when epoch =1. 
In cases where not all unique pairs were generated in epoch 1 they will now continue to be draw new samples in later epochs.
- This is quite a major code change for quite a rare case and again, the benefits are only seem at very low `num_iterations` and epochs > 1. I personally like the democratisation with the sentence-transformers "samplers" though, which of course SetFit is very closely linked to.

Thoughts as ever appreciated @tomaarsen on this

tomaarsen:
I'm a great fan of how this is looking! I don't have a lot of time to dive into it now (I'll look on Monday), but is my initial assumption correct that with these changes in place, increasing `num_epochs` would become essentially equivalent to increasing `num_iterations`?
I think we make the entire project a lot more intuitive if `num_iterations` is deprecated.

tomaarsen:
I ran some tests with the new `ContrastiveDataset`: As expected, increasing `num_epochs` is now exactly equivalent to increasing `num_iterations`. To be concrete, the following situations sample exactly the same pairs:
* `num_epochs = 1`, `num_iterations = 20`
* `num_epochs = 2`, `num_iterations = 10`
* `num_epochs = 20`, `num_iterations = 1`

This affords us to deprecate one of them (i.e. `num_iterations`), which may help simplify the code further, too. 
An interesting implementation may be to create two generators: for positive pairs and for negative pairs. These generators can be "refreshed" once they "run out". I think we can then also simplify by only offering unique pairs per epoch, which means the `unique_pairs` parameter could perhaps be removed as well. 
If beneficial, we could introduce an `even_sampling` or `sample_evenly` argument that decides whether the number of positive and negative pairs should be equal. Alternatively, we can provide e.g. `sampling_strategy` which accepts either `"oversampling"`, `"undersampling"` or e.g. `"unbalanced"`.

I'd love to know your thoughts on this.

- Tom Aarsen

vahuja4:
The experiments here look very interesting and promising. @tomaarsen - would you please tell me if these changes are going to be merged soon?

tomaarsen:
This has been superseded by https://github.com/tomaarsen/setfit/pull/5, which has been merged into [`v1.0.0-pre`](https://github.com/huggingface/setfit/tree/v1.0.0-pre). I'll close this accordingly. Thanks for the great work @danstan5!

- Tom Aarsen

PR #272, opened 2023-01-14
Feature/setfithead multi target

Yongtae723:
I tried to solve conflict error and fix some error 

please check this PR is what you intended.

also, some I failed some test...and I cannot figure out the reason...

### DIFF
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -278,5 +278,5 @@ class SetFitDataset(TorchDataset):
         # convert to tensors
         features = {k: torch.Tensor(v).int() for k, v in features.items()}
         labels = torch.Tensor(labels)
-        labels = labels.long() if len(labels.size())== 1 else labels.float()
+        labels = labels.long() if len(labels.size()) == 1 else labels.float()
         return features, labels


tests/test_trainer.py
```diff
@@ -285,6 +285,53 @@ def compute_metrics(y_pred, y_test):
...
+        trainer = SetFitTrainer(
+            model=self.model,
+            train_dataset=dataset,
+            eval_dataset=dataset,
+            metric=compute_metrics,
+            num_iterations=self.num_iterations,
+            column_mapping={"text_new": "text", "label_new": "label"},
+        )
+
+        # trainer.freeze()
```
Yongtae723, line 319: @tomaarsen 
If I remove comment out, an error occurs.

sorry I don't know the reason. do you have some idea?

tests/test_trainer.py
```diff
@@ -285,6 +285,53 @@ def compute_metrics(y_pred, y_test):
...
+        trainer.train()
+
+        # trainer.unfreeze(keep_body_frozen=False)
+        trainer.train(5)
+        metrics = trainer.evaluate()
+
+        self.assertEqual(
```
Yongtae723, line 326: I dont know why, but metrics is not want you intended...

src/setfit/modeling.py
```diff
@@ -179,12 +186,12 @@ def forward(
...
-
         x = features["sentence_embedding"] if is_features_dict else features
         logits = self.linear(x)
-        logits = logits / (temperature + self.eps)
-        probs = nn.functional.softmax(logits, dim=-1)
-
+        if self.multitarget or self.out_features == 1:  # only has one target or multiple targets per item
+            probs = torch.sigmoid(logits / temperature)
+        else:  # multiple classes, one target per item
+            probs = nn.functional.softmax(logits / temperature, dim=-1)
```
tomaarsen, line 194: `self.eps` is not used here

src/setfit/modeling.py
```diff
@@ -179,12 +186,12 @@ def forward(
         if isinstance(features, dict):
             assert "sentence_embedding" in features
             is_features_dict = True
-
         x = features["sentence_embedding"] if is_features_dict else features
         logits = self.linear(x)
-        logits = logits / (temperature + self.eps)
-        probs = nn.functional.softmax(logits, dim=-1)
-
+        if self.multitarget or self.out_features == 1:  # only has one target or multiple targets per item
```
tomaarsen, line 191: `self.out_features == 1` is no longer used, see:
https://github.com/huggingface/setfit/blob/f777c2c60b270604dae0dc1db4eea815e8c9019d/src/setfit/modeling.py#L135-L139

src/setfit/modeling.py
```diff
@@ -196,20 +203,32 @@ def forward(
...
+        is_tensor = isinstance(x_test, torch.Tensor)  # Otherwise assume it's ndarray
+        if not is_tensor:
+            x_test = torch.Tensor(x_test).to(self.device)
         self.eval()
 
-        return self(x_test)[1]
+        out = self(x_test)[1]
+        if not is_tensor:
+            return out.detach().cpu().numpy()
+        return out
```
tomaarsen, line 215: We deprecated `np.ndarray` support for `SetFitHead` recently. We now rely on `SetFitModel` to always supply the Head with torch Tensors.

src/setfit/modeling.py
```diff
@@ -196,20 +203,32 @@ def forward(
...
 
-    def predict(self, x_test: torch.Tensor) -> torch.Tensor:
+    def predict(self, x_test: Union[torch.Tensor, "ndarray"]) -> Union[torch.Tensor, "ndarray"]:
         probs = self.predict_proba(x_test)
 
-        out = torch.argmax(probs, dim=-1)
-
+        if self.out_features == 1 or self.multitarget:
```
tomaarsen, line 220: `self.out_features == 1` is no longer used.

src/setfit/modeling.py
```diff
@@ -196,20 +203,32 @@ def forward(
...
+            out = np.where(probs >= 0.5, 1, 0) if isinstance(probs, np.ndarray) else torch.where(probs >= 0.5, 1, 0)
+            # TODO 0.5 is not suitable. I will set this as threshold in Next PR.
+        else:
+            out = np.argmax(probs, dim=-1) if isinstance(probs, np.ndarray) else torch.argmax(probs, dim=-1)
```
tomaarsen, line 224: We don't need to have `np` support here anymore.

src/setfit/modeling.py
```diff
@@ -196,20 +203,32 @@ def forward(
...
         return out
 
     def get_loss_fn(self):
-        return torch.nn.CrossEntropyLoss()
+        if self.out_features == 1 or self.multitarget:  # if sigmoid output
```
tomaarsen, line 228: `self.out_features == 1` is no longer used.

src/setfit/modeling.py
```diff
@@ -404,7 +425,7 @@ def predict_proba(self, x_test: List[str], as_numpy: bool = False) -> Union[torc
         outputs = self.model_head.predict_proba(embeddings)
 
         if as_numpy and self.has_differentiable_head:
-            outputs = outputs.cpu().numpy()
+            outputs = outputs.cpu().detach().numpy()
```
tomaarsen, line 428: Good call! I believe `outputs.detach().cpu().numpy()` is more efficient, so I'll go with that. I'll also apply it for `predict`.

tomaarsen (commented):
Thanks for these changes! I think we're getting really close. I made some comments based on some recent PRs that got merged since #212 was made. In particular, removing support for numpy arrays in the differentiable head, and removing `out_features=1`: 2 is now the minimum. 

I've made these changes and pushed them to this PR. Make sure to `git pull` them if you want to make more changes of your own. In short, all of the comments that I made with this review are now resolved.

As for your comments regarding the `trainer.freeze()`, I'm not sure what caused the issue, but it seems to be gone after I made my changes.

tomaarsen:
I ran some experiments using the multiclass classification for the different heads.

## Dataset

<details><summary>Dataset generation script</summary>

```python

dataset = load_dataset("SetFit/hate_speech_offensive")


def to_multiclass(sample):
    """
    from
        (0: 'hate-speech', 1: 'offensive-language' or 2: 'neither')
    to
        ([1, 0]: 'hate-speech', [1, 1]: 'offensive-language' or [0, 0]: 'neither')
    """
    label = sample["label"]
    sample["label"] = [1 if label == 0 else 0, 1 if label == 1 else 0]
    return sample


train_dataset = sample_dataset(dataset["train"], label_column="label", num_samples=8).map(to_multiclass)
eval_dataset = dataset["test"].map(to_multiclass)
```
</details>

I want to point out that this isn't a very natural use of a multiclass dataset. That said, I couldn't find an actual multiclass dataset on the Hub. 

## Training Scripts
<details><summary>Logistic Regression head</summary>

```python
trainer = SetFitTrainer(
    model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
)
trainer.train()
metrics = trainer.evaluate()
```
</details>

<details><summary>Differentiable head</summary>

```python
trainer = SetFitTrainer(
    model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
)

# Train and evaluate
trainer.freeze() # Freeze the head
trainer.train() # Train only the body

# Unfreeze the head and freeze the body -> head-only training

trainer.train(
    num_epochs=25, # The number of epochs to train the head or the whole model (body and head)
    batch_size=16,
    body_learning_rate=1e-5, # The body's learning rate
    learning_rate=1e-2, # The head's learning rate
    l2_weight=0.0, # Weight decay on **both** the body and head. If `None`, will use 0.01.
)
```
</details>

And the model for testing was `"sentence-transformers/paraphrase-mpnet-base-v2"`.

## Results
| Model                                               | Evaluation Accuracy   |
|-----------------------------------------------------|-----------------|
| Multilabel Logistic Regression ("multi-output")     | 0.5550 (0.0000) |
| Multilabel Logistic Regression ("one-vs-rest")      | 0.5550 (0.0000) |
| Multilabel Logistic Regression ("classifier-chain") | 0.5580 (0.0000) |
| Multilabel Differentiable Head                      | 0.5849 (0.0044) |

Notes:
1. Evaluation accuracy is displayed as mean and standard deviation between 5 executions.
2. The same seed is used in `SetFitTrainer` and `sample_dataset` between all executions, so all executions work on the same training data.
3. The Logistic Regression performance was identical between different executions (perhaps due to the same seed between executions)
4. For the differentiable head, there is no difference between the "multi-output" and "one-vs-rest" strategies, and "classifier-chain" is unsupported. So, I ran experiments only with "multi-output".

To me, this is indicative that the multilabel differentiable head performs equivalently to the logistic regression head. With other words, this PR seems to be successful in adding support to SetFitHead for with multi-label classification! 🎉 

- Tom Aarsen

Yongtae723:
I always feel thank you for your thoughtful comments and edits. (Also comment of previous Issues)

I will check your editing!
thank you!

Yongtae723:
@tomaarsen 
I also think the readme should be edited.

I think it would be better to submit the edit of the readme in a separate PR.
how do you think?

Yongtae723:
I confirmed your change! thank you!

Also, I did a similar multi-label example and get a similar result.

So I think you can merge it into the main!

thank you

tomaarsen:
> @tomaarsen I also think the readme should be edited.
> 
> I think it would be better to submit the edit of the readme in a separate PR. how do you think?

If the changes that you have planned to the README related to the changes from this PR, then I think the changes should be included in this PR. That way, the code and README get updated at the same time. I'm glad to hear that your experiments work too!

Yongtae723:
I got it!

I edited the readme to reflect our changes.
Since I am not a native English speaker, my written English might be strange.
So please rewrite the readme if my English is not correct.

src/setfit/modeling.py
```diff
@@ -295,6 +305,8 @@ def fit(
                     # to model's device
                     features = {k: v.to(device) for k, v in features.items()}
                     labels = labels.to(device)
+                    if self.model_head.multitarget:
+                        labels = labels.float()
```
tomaarsen, line 309: My only gripe right now is this section. In the SetFitDataset we explicitly convert the labels to `long`, as that is required for the `CrossEntropyLoss`: 
https://github.com/huggingface/setfit/blob/4fe132a2ed1b60c37acc9888021ee692cdb03ac1/src/setfit/data.py#L280

However, if multitarget is set, we *again* cast the labels, now to floats. I think we should try to modify the `SetFitDataset` such that it returns the correctly typed label from the get-go, rather than hackishly having to cast them to float mid-training.

tomaarsen (changes requested):
I'm satisfied with almost everything in this PR, with one exception. I'm not sure what the best course of action is, nor what the "normal" approach for this is. Perhaps to provide the `SetFitDataset` with a `label_postprocessing` function that either converts them to floats or to longs, depending on what is needed?

Yongtae723:
I understand your concirm and agree with you.

I fix some code for what you want to do.

src/setfit/data.py
```diff
@@ -277,6 +277,7 @@ def collate_fn(batch):
 
         # convert to tensors
         features = {k: torch.Tensor(v).int() for k, v in features.items()}
-        labels = torch.Tensor(labels).long()
+        labels = torch.Tensor(labels)
+        labels = labels.long() if isinstance(label, int) else labels.float()
```
Yongtae723, line 281: my suggestion is to use a type of label

The type of label should be 'int' for the single label classification, but 'List' for multilabel classification.

we can write
```
label =  torch.Tensor(labels).long() if isinstance(label, int) else  torch.Tensor(labels).float()
```
but I felt it is too long so that I fix as I push.

Yongtae723: Or
```python
labels = labels.long() if len(labels[0])== 1 else labels.float()
```

whichever you want!

tomaarsen: I think the second solution is best! 
```python
labels = labels.long() if len(labels.size()) == 1 else labels.float()
```
That should accurately measure whether we are in a multitarget situation, even if the user accidentally supplies floats instead of integers.

Yongtae723: Oh, I see.

your suggestion makes sense to me!
I will fix that!

tomaarsen (approved):
Looks good to me now! Thanks for making all of these changes! 🎉

PR #280, opened 2023-01-18
Fix squared optimization steps bug

twerkmeister:
Currently, when increasing the number of epochs, the number of steps per epoch is also increased, leading to a number of optimization steps thats roughly `(examples / batch size) * epochs * epochs` instead of `(examples / batch size) * epochs`. The proposed change should fix this.

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -361,7 +361,7 @@ class SetFitTrainer:
 
                 train_dataloader = DataLoader(train_examples, shuffle=True, batch_size=batch_size)
                 train_loss = self.loss_class(self.model.model_body)
-      
+
             total_train_steps = len(train_dataloader) * num_epochs
             logger.info("***** Running training *****")
             logger.info(f"  Num examples = {len(train_examples)}")


tomaarsen:
Hello!

This bug was discovered https://github.com/huggingface/setfit/pull/268#issuecomment-1382924584, and a fix was proposed. It involves simply removing this line:
https://github.com/huggingface/setfit/blob/f777c2c60b270604dae0dc1db4eea815e8c9019d/src/setfit/trainer.py#L378
When `steps_per_epoch` is not specified, the number of steps per epoch will be equivalent to the number of datapoints in the dataloader. See the [docs](https://sbert.net/docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.fit) on the `fit` method for more info.

The remainder of the code would not need any changes.
Would you agree that this is a preferable solution over the changes that you have proposed? I think it's best if we make a new PR for this.

- Tom Aarsen

twerkmeister:
Hey @tomaarsen, I removed the argument as you mentioned. Beyond that, this version also fixes the fact that for the triplet losses `self.num_epochs` was used to calculate total number of optimization steps for reporting when in fact `num_epochs` should be used because `self.num_epochs` can be shadowed by an argument to the train method

tomaarsen:
You're very right! Thanks for catching that. I'll merge this once the tests go green 🎉

src/setfit/trainer.py
```diff
@@ -363,19 +361,17 @@ def train(
...
-                train_steps = len(train_dataloader) * num_epochs
 
             logger.info("***** Running training *****")
             logger.info(f"  Num examples = {len(train_examples)}")
             logger.info(f"  Num epochs = {num_epochs}")
-            logger.info(f"  Total optimization steps = {train_steps}")
+            logger.info(f"  Total optimization steps = {len(train_dataloader) * num_epochs}")
             logger.info(f"  Total train batch size = {batch_size}")
 
             warmup_steps = math.ceil(train_steps * self.warmup_proportion)
```
tomaarsen, line 371:
```suggestion
            warmup_steps = math.ceil(train_steps * self.warmup_proportion)
```
This still uses the `train_steps` variable.

tomaarsen:
Wonderful! Thank you for this fix, and for the editions on the PR @twerkmeister!

PR #283, opened 2023-01-22
Resolve `SentenceTransformer` resetting devices after moving a `SetFitModel`

tomaarsen:
Hello!

## Pull Request overview
* Resolve bug that moved a `SentenceTransformer` instance back to its original device after `encode` or `fit` is called on it.
* Add passing test that fails on main to prevent this bug from reappearing.

## Note
I want to first point out that this PR is cherry-picked from a bugfix that I applied in #265. I believe it is best to merge that PR to create a v1.0.0 version, and merge this PR in preparation for a v0.6.0 version. That way, we'll have a version out there without breaking changes that includes some important bug fixes (i.e. v0.6.0). 
However, if we choose to include #265 in the next release already, then that PR would supersede this one, and this one could be closed.

## Details
The bug in question is most conveniently showcased with some code:
```python
import torch

print(f"{torch.cuda.is_available()=}")

model: SetFitModel = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
print(f"{model.model_body.device=} after loading")

model.to("cpu")
print(f"{model.model_body.device=} after moving to cpu")

model.model_body.encode("Arbitrary sentence to convert to embeddings")
print(f"{model.model_body.device=} after encoding some inputs")
```
With the following outputs on a CUDA-supported device:
```
torch.cuda.is_available()=True
model.model_body.device=device(type='cuda', index=0) after loading
model.model_body.device=device(type='cpu') after moving to cpu
model.model_body.device=device(type='cuda', index=0) after encoding some inputs
```
As you can see, after calling `model.model_body.encode`, the `SentenceTransformer` has moved back to `cuda:0`.

This is a bug/flaw originating from the SentenceTransformer library. When initializing a ST, it will set an attribute `self._target_device = device` [here](https://github.com/UKPLab/sentence-transformers/blob/b86eec31cf0a102ad786ba1ff31bfeb4998d3ca5/sentence_transformers/SentenceTransformer.py#L107), and then it will continuously move the ST to that original device in the [encode](https://github.com/UKPLab/sentence-transformers/blob/b86eec31cf0a102ad786ba1ff31bfeb4998d3ca5/sentence_transformers/SentenceTransformer.py#L150-L151) and [fit](https://github.com/UKPLab/sentence-transformers/blob/b86eec31cf0a102ad786ba1ff31bfeb4998d3ca5/sentence_transformers/SentenceTransformer.py#L641) methods. This would be fine if calling `to` on such an ST would update this target device, but it does **not**.

I would create a PR over on the [SentenceTransformer repository](https://github.com/UKPLab/sentence-transformers), but I'm afraid that repository hasn't seen any maintenance is 2 months, with issues and PRs piling up. As a result, I believe it's prudent to solve this on the SetFit side, sadly.

## Tests
I've created a simple test very much alike the snippet from above. It will only run on CUDA-enabled devices, as with CPU-only devices there is no second device to move a model to.

- Tom Aarsen

### DIFF
--- a/tests/test_modeling.py
+++ b/tests/test_modeling.py
@@ -262,10 +262,13 @@ def test_to_torch_head():
 
 
 @torch_cuda_available
-def test_to_sentence_transformer_device_reset():
+@pytest.mark.parametrize("use_differentiable_head", [True, False])
+def test_to_sentence_transformer_device_reset(use_differentiable_head):
     # This should initialize SentenceTransformer() without a specific device
     # which sets the model to CUDA iff CUDA is available.
-    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
+    model = SetFitModel.from_pretrained(
+        "sentence-transformers/paraphrase-albert-small-v2", use_differentiable_head=use_differentiable_head
+    )
     # If we move the entire model to CPU, we expect it to stay on CPU forever,
     # Even after encoding or fitting
     model.to("cpu")


lewtun (approved):
Thanks for this fix and super clean explanation in the PR description @tomaarsen ! I agree we can merge this and release a `v0.6.0` before merging #265 for a `v1.0.0` release

tomaarsen:
Great!

Issue #287, opened 2023-01-24
Problem using distiluse models: forward() got an unexpected keyword argument 'token_type_ids' - possible fix


AleksanderObuchowski:
I have the following error while running SetFit with distiluse models (e.g. [this one](https://huggingface.co/sentence-transformers/distiluse-base-multilingual-cased-v1)):
When I try to fit the model I get

`TypeError: forward() got an unexpected keyword argument 'token_type_ids'`

From what I can see looking at the SetFitDataset the tokenizer always return the `token_type_ids` even if the Sentence Transformers model does not support them.

``` 
def __getitem__(self, idx: int) -> Tuple[TokenizerOutput, Union[int, List[int]]]:
        feature = self.tokenizer(
            self.x[idx],
            max_length=self.max_length,
            padding="max_length",
            truncation=True,
            return_attention_mask=True,
            return_token_type_ids=True,
        )
        label = self.y[idx]

        return feature, label
```
I think this could be changed to something like:

```
def __getitem__(self, idx: int) -> Tuple[TokenizerOutput, Union[int, List[int]]]:
        feature = self.tokenizer(
            self.x[idx],
            max_length=self.max_length,
            padding="max_length",
            truncation=True,
            return_attention_mask=True,
            return_token_type_ids=True if "token_type_ids" in self.tokenizer.model_input_names else False,
        )
        label = self.y[idx]

        return feature, label

```

And similarly, the `collate_fn` could be changed to output `token_type_ids` only if the model input expects it based on the `tokenizer.model_input_names`

I will be happy to make the PR if you think this is the right approach.

tomaarsen:
Thank you for reporting! We would happily welcome your PR. To help out out already, I believe your solution can be simplified from:
```python
return_token_type_ids=True if "token_type_ids" in self.tokenizer.model_input_names else False,
```
to
```python
return_token_type_ids="token_type_ids" in self.tokenizer.model_input_names,
```

Furthermore, I would appreciate it if you could expand this fix to `return_attention_mask`, too. I assume most models require `attention_mask`, but some extra robustness is always appreciated.

- Tom Aarsen

AleksanderObuchowski:
Thanks for the simplification, I created a PR https://github.com/huggingface/setfit/pull/288, I also changed the `collate_fn` form static method to normal method of the dataset as now the collator has to know what tokenizer is being used. I think it shouldn't cause any issues.

harinisri2001:
HI @AleksanderObuchowski 
[...]

PR #288, opened 2023-01-24
Dynamic features in datasets based on model input names

AleksanderObuchowski:
As mentioned in https://github.com/huggingface/setfit/issues/287  hard-coded feature names cause problems with some models not using certain features such as "token_type_ids" so now the feature are dynamic based on the `model_input_names` parameter of tokenizer.

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -4,10 +4,13 @@ import numpy as np
 import pandas as pd
 import pytest
 from datasets import Dataset, load_dataset
+from torch.utils.data import DataLoader
+from transformers import AutoTokenizer
 
 from setfit.data import (
     SAMPLE_SIZES,
     SEEDS,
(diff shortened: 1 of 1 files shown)


tomaarsen:
I've added a simple test that verifies that the batch data has the right parameters for the model, and have included the `sentence-transformers/distiluse-base-multilingual-cased-v1` model that you mentioned, as well as a model more frequently used throughout the tests. As expected, the test fails for the `main` branch due to the bug that this PR fixes.

Do you know of any other models that may be worth testing? 

If not, then I think this is ready to merge when the tests go green.

AleksanderObuchowski:
I can't think of any other models right now

tomaarsen:
No problem, me neither. Let's just merge this then. Thanks for the bug report and the PR @AleksanderObuchowski!

PR #292, opened 2023-01-30
add `run_zeroshot.py`; add functionality to `data.get_templated_dataset()` (formerly `add_templated_examples()`)

danielkorat:
This PR simplifies running zero shot with SetFit. 
The goal is to add a `run_zeroshot.py` script and remove the function `data.get_augmented_samples()` since its functionality is included in the generic `data.get_templated_dataset()`.

### DIFF
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -110,7 +110,7 @@ def get_templated_dataset(
     return dataset
 
 
-def add_templated_examples(**kwargs) -> None:
+def add_templated_examples(*args, **kwargs) -> None:
     warnings.warn(
         "`add_templated_examples` has been deprecated and will be removed in v1.0.0 of SetFit. "
         " Please use `get_templated_dataset` instead.",


danielkorat:
Still work in progress

tomaarsen:
This section:
https://github.com/huggingface/setfit/blob/5e78e40933e53fac85efeff666818287bd52237b/tests/test_data.py#L177-L189

is a leftover decorator from the now-removed `test_get_augmented_samples`. The decorator should also be removed if you want to remove `test_get_augmented_samples`.

danielkorat:
> @danielkorat This section:
> 
> https://github.com/huggingface/setfit/blob/5e78e40933e53fac85efeff666818287bd52237b/tests/test_data.py#L177-L189
> 
> is a leftover decorator from the now-removed `test_get_augmented_samples`. The decorator should also be removed if you want to remove `test_get_augmented_samples`.

@tomaarsen 
I re-implemented these tests now.

src/setfit/data.py
```diff
@@ -1,8 +1,8 @@
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Tuple, Union
```
tomaarsen, line 1:
```suggestion
from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
```
Required for recommended changes below:

src/setfit/data.py
```diff
@@ -15,69 +15,115 @@
...
-            "label": [0] * sample_size + [1] * sample_size + [2] * sample_size,
-        }
-    elif dataset == "ade_corpus_v2_classification":
-        return {
-            "text": ["The sentence is not related"] * sample_size + ["The sentence is related"] * sample_size,
-            "label": [0] * sample_size + [1] * sample_size,
+def get_templated_dataset(
+    dataset: Dataset = None,
+    candidate_labels: List[str] = None,
+    reference_dataset: str = None,
```
tomaarsen, line 21:
```suggestion
    dataset: Optional[Dataset] = None,
    candidate_labels: Optional[List[str]] = None,
    reference_dataset: Optional[str] = None,
```
Adding `Optional` in accordance to [PEP 484](https://peps.python.org/pep-0484/#union-types), I quote:
> A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
> ```
> def handle_employee(e: Employee = None): ...
> ```
> This would have been treated as equivalent to:
> ```
> def handle_employee(e: Optional[Employee] = None) -> None: ...
> ```
> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

src/setfit/data.py
```diff
@@ -15,69 +15,115 @@
...
+    template: str = "This sentence is {}",
+    sample_size: int = 2,
+    text_column: str = "text",
+    label_column: str = "label",
+    multi_label: bool = False,
+    label_names_column: str = "label_text",
+) -> Dataset:
+    """Create templated examples for a reference dataset or refernece labels.
```
tomaarsen, line 29:
```suggestion
    """Create templated examples for a reference dataset or reference labels.
```
Typo

src/setfit/data.py
```diff
@@ -15,69 +15,115 @@
...
+    If input Dataset is supplied, add the examples to it, otherwise create a new Dataset.
+    The input Dataset is assumed to have a text column with the name `text_column` and a
+    label column with the name `label_column`, which contains one-hot or multi-hot
+    encoded label sequences.
+
+    Args:
+        dataset (`Dataset`): A Dataset to add templated examples to.
+        candidate_labels (`List[str]`, defaults to None): The list of candidate
+        labels to be fed into the template to construct examples.
+        reference_dataset (`str`, defaults to None): A dataset to take labels
```
tomaarsen, line 43:
```suggestion
        dataset (`Dataset`, *optional*): A Dataset to add templated examples to.
        candidate_labels (`List[str]`, *optional*): The list of candidate
        labels to be fed into the template to construct examples.
        reference_dataset (`str`, *optional*): A dataset to take labels
```
I'm not 100% sure on the recommend Google docstring syntax, but I think this is it?

src/setfit/data.py
```diff
@@ -15,69 +15,115 @@
...
+        # Some datasets on the Hugging Face Hub don't have a ClassLabel feature for the label column.
+        # In these cases, you should compute the candidate labels manually by first computing the id2label mapping.
+
+        # The column with the label names
+        label_names = dataset.unique(label_names_column)
+        # The column with the label IDs
+        label_ids = dataset.unique("label")
+        id2label = dict(zip(label_ids, label_names))
+        # Sort by label ID
+        id2label_sorted = {key: val for key, val in sorted(id2label.items(), key=lambda x: x[0])}
```
tomaarsen, line 122: Can this not be simplified to something like
```python
id2label_sorted = dict(sorted(id2label.items(), key=lambda x: x[0]))
```
I can't test this right now, but I think it's possible. 

Also, you can probably just
```python
_, candidate_labels = zip(*sorted(id2label.items(), key=lambda x: x[0]))
```
But that's also off the top of my head, not 100% sure if that'd work.

src/setfit/data.py
```diff
@@ -110,21 +156,24 @@ def create_fewshot_splits(
     dataset: Dataset,
     sample_sizes: List[int],
     add_data_augmentation: bool = False,
-    dataset_name: Optional[str] = None,
+    dataset_name: str = None,
```
tomaarsen, line 159:
```suggestion
    dataset_name: Optional[str] = None,
```
[PEP 484](https://peps.python.org/pep-0484/#union-types) again, if `None` is the default, then the type hint must include `Optional`.

tomaarsen (changes requested):
I have some small nitpicks, could you have a look at them?

src/setfit/data.py
danielkorat: Thanks @tomaarsen for all your important comments  :)
I simplified it without using `dict`.
Ready for review.

tomaarsen (approved):
Looks good to me! `get_templated_dataset` feels pretty intuitive to use.
I'll let @lewtun decide if we need to keep `add_templated_examples` and `add_augmented_samples`, which this PR removes, but fully replace the function bodies with a deprecation warning instead. These warnings can point to the new `get_templated_dataset`. I suspect that the code of some notebooks or tutorials out there might break without explanation otherwise.

danielkorat:
> Looks good to me! `get_templated_dataset` feels pretty intuitive to use. I'll let @lewtun decide if we need to keep `add_templated_examples` and `add_augmented_samples`, which this PR removes, but fully replace the function bodies with a deprecation warning instead. These warnings can point to the new `get_templated_dataset`. I suspect that the code of some notebooks or tutorials out there might break without explanation otherwise.

`get_templated_dataset` is backwards compatible with `add_templated_examples`. So it's just a name change. `add_templated_examples` was only used in tests and in the `zero-shot-classification.ipynb` notebook. I refactored these files accordingly, so there is no mention of `add_templated_examples` anymore. We can add a line like:

`add_templated_examples = get_templated_dataset`

with a deprecation warning. What do you think @tomaarsen ?

tomaarsen:
A deprecation warning would be great. Could you include a suggestion of the new function to use as well as that the old function will be removed in `v1.0.0`?

src/setfit/data.py
```diff
@@ -15,69 +16,124 @@
...
         }
-    else:
-        raise ValueError(f"Dataset {dataset} not supported for data augmentation!")
+        for _ in range(sample_size):
+            dataset = dataset.add_item(example)
+
+    return dataset
+
+
+def add_templated_examples(**kwargs) -> None:
```
tomaarsen, line 104:
```suggestion
def add_templated_examples(*args, **kwargs) -> None:
```
I think there's an error if users call `add_templated_examples(dataset)` otherwise.

src/setfit/data.py
```diff
@@ -15,69 +16,124 @@
...
+    warnings.warn(
+        "`add_templated_examples` has been deprecated and will be removed in v1.0.0 of SetFit. "
+        " Please use `get_templated_dataset` instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
```
tomaarsen, line 111: Can we add a near-identical copy for `get_augmented_samples` which also recommends to use `get_templated_dataset`? That way, everyone's code will at least give them a warning pointing them to what they should be doing for the new version (even if few people might be using `get_augmented_samples`).

tomaarsen (changes requested):
I have some tiny last nitpicks

src/setfit/data.py
danielkorat: thanks

src/setfit/data.py
danielkorat: thanks

src/setfit/data.py
danielkorat: Hi @tomaarsen not sure I understand. I already pushed an empty `get_augmented_samples` with the warning. Please clarify or suggest code

danielkorat: What do you mean by "identical" ? thanks

tomaarsen: I'm referring to this method: https://github.com/huggingface/setfit/blob/aa07a854fcfb2605dd809c19cbf1919015873d13/src/setfit/data.py#L18
This PR removes it, and I think it would be best if we add:
```python
def get_augmented_samples(*args, **kwargs) -> None:
    warnings.warn(
        "`get_augmented_samples` has been deprecated and will be removed in v1.0.0 of SetFit. "
        "Please use `get_templated_dataset` instead.",
        DeprecationWarning,
        stacklevel=2,
    )
```

danielkorat: Oh I see. sorry for the mix-up. Will add :)

danielkorat:
@tomaarsen please review

tomaarsen:
Thank you @danielkorat for sticking with me through my many suggestions. The script is very useful, as is the new and improved `get_templated_dataset()`. I'm very impressed at the results of zeroshot as well: it is very consistent in its results compared to 8-shot, and zeroshot seems to outperform 8-shot in various benchmarks (e.g. `SetFit/emotion`).

Issue #226, opened 2022-12-10
ValueError: A column mapping must be provided when the dataset does not contain the following columns: {'text', 'label'}


Cheril184:
I am getting  this value error in trainer.evaluate() despite my dataset containing columns named 'text' and 'label'

my dataset dict looks like this:
Dataset({
    features: ['text', 'reason', 'label'],
    num_rows: 2062
})

Cheril184:
@LuketheDukeBates can you please help?

theainerd:
Hello @Cheril184,

Map the features like this. Use the `column_mapping` to map with the column names.

```python
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    loss_class=CosineSimilarityLoss,
    num_iterations=20,
    column_mapping={"text": "text", "label": "label"},
)
```

tomaarsen:
If the datasets contain the `text` and `label` columns, then a column mapping should not be needed.

I'm unable to reproduce your issue. I'm using the following dummy script in my attempt:
```python
from setfit import SetFitModel, SetFitTrainer
from datasets import Dataset

ds = Dataset.from_dict({"text": ["a", "b", "c"], "reason": ["d", "e", "f"], "label": [0, 1, 2]})
print(ds)

model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")

trainer = SetFitTrainer(
    model=model,
    train_dataset=ds,
    eval_dataset=ds,
    num_iterations=20,
)

trainer.train()
print(trainer.evaluate())
```
This outputs
```
Dataset({
    features: ['text', 'reason', 'label'],
    num_rows: 3
})
model_head.pkl not found on HuggingFace Hub, initialising classification head with random weights. You should TRAIN this model on a downstream task to use it for predictions and inference.
***** Running training *****
  Num examples = 120
  Num epochs = 1
  Total optimization steps = 8
  Total train batch size = 16
Epoch:   0%|                                                                                                                                                                    | 0/1 [00:00<?, ?it/s]
Iteration:   0%|                                                                                                                                                                | 0/8 [00:00<?, ?it/s] 
Iteration:  12%|███████████████████                                                                                                                                     | 1/8 [00:02<00:15,  2.27s/it]
Iteration:  25%|██████████████████████████████████████                                                                                                                  | 2/8 [00:02<00:07,  1.28s/it]
[...]

PR #299, opened 2023-02-06
Add a warning if an unsplit dataset is passed to SetFitTrainer

jaalu:
Adds a warning for a potential cause of issue #226.

This isn't really a problem in the library as such, but Huggingface Datasets' default behavior (as of `datasets 2.9.0`) when loading .CSV or .JSON files through `load_dataset` is to automatically assign the rows to the training split:
```
import datasets
with open('example.csv', 'w') as example:
    example.writelines(["label,text\n","1,good\n","0,terrible\n"])
features = datasets.Features(
    { 
    'label': datasets.ClassLabel(num_classes=2, names=['positive','negative']),
    'text': datasets.Value('string')
    })
dataset = datasets.load_dataset("csv", data_files="./example.csv", features=features)
# Prints
# DatasetDict({
#     train: Dataset({
#         features: ['label', 'text'],
#         num_rows: 2
#     })
# })
print(dataset)
```

As a result, the "obvious" way to quickly train a classifier
```
import datasets

with open('example.csv', 'w') as example:
    example.writelines(["label,text\n","1,good\n","0,terrible\n"])
# ERROR: We forget to set a split here:
ds = datasets.load_dataset("csv", data_files="./example.csv")

model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")

trainer = SetFitTrainer(
    model=model,
    train_dataset=ds,
    eval_dataset=ds,
    num_iterations=20,
)

trainer.train()
```

fails with the error
```
  File "C:\Projects\setfit\splitexample.py", line 17, in <module>
    trainer.train()
  File "c:\projects\setfit\src\setfit\trainer.py", line 310, in train
    self._validate_column_mapping(self.train_dataset)
  File "c:\projects\setfit\src\setfit\trainer.py", line 145, in _validate_column_mapping
    raise ValueError(
ValueError: A column mapping must be provided when the dataset does not contain the following columns: {'text', 'label'}
```

which suggests an error in the formatting of the data (rather than the split). Warning on this case should hopefully make it easier to spot the core problem.

### Possible alternatives

* The CSV loading script in Datasets itself could warn that a split has been picked automatically, but it's hard to tell how much code relies on the existing behavior. 
* Detecting and selecting the appropriate split automatically would be nicer for the end user, but would also be hard to undo later if the Datasets or the library API changes.

### DIFF
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -1,9 +1,11 @@
+import pathlib
 import re
+import tempfile
 from unittest import TestCase
 
 import evaluate
 import pytest
-from datasets import Dataset
+from datasets import Dataset, load_dataset
 from sentence_transformers import losses
 from transformers.testing_utils import require_optuna
 from transformers.utils.hp_naming import TrialShortNamer
@@ -116,6 +118,22 @@ class SetFitTrainerTest(TestCase):
         expected_message = re.escape(
             "SetFit expected a Dataset, but it got a DatasetDict with the splits ['test', 'train']. "
             "Did you mean to select one of these splits from the dataset?",
+        )
+        with pytest.raises(ValueError, match=expected_message):
+            trainer._validate_column_mapping(trainer.train_dataset)
+
+    def test_trainer_raises_error_when_dataset_is_dataset_dict_with_train(self):
+        """Verify that a useful error is raised if we pass an unsplit dataset with only a `train` split to the trainer."""
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            path = pathlib.Path(tmpdirname) / "test_dataset_dict_with_train.csv"
+            path.write_text("label,text\n1,good\n0,terrible\n")
+            dataset = load_dataset("csv", data_files=str(path))
+        trainer = SetFitTrainer(
+            model=self.model, train_dataset=dataset, eval_dataset=dataset, num_iterations=self.num_iterations
+        )
+        expected_message = re.escape(
+            "SetFit expected a Dataset, but it got a DatasetDict with the split ['train']. "
+            "Did you mean to select the training split with dataset['train']?",
         )
         with pytest.raises(ValueError, match=expected_message):
             trainer._validate_column_mapping(trainer.train_dataset)


jaalu:
> I think we might be able to make the conditional even more robust if we perform an if isinstance(dataset, DatasetDict). That way, we don't have to guess the user's split names (train, test, dev, eval, etc.).

I think that's a very good idea! `isinstance(dataset, DatasetDict)` works for the cases I've tested and makes the intent clearer. I'm still inclined towards having a special case for when you just have a single `train` column, since it's the case that's easiest to unintentionally run into, but I've added this in a separate commit so it's trivial to revert.

> Alternatively, we can log the columns that we do find, rather than only writing an error message with the ones that we can't. After all, that would also give users valuable insight in what might be going wrong.

I don't think this is an either-or - logging the column names definitely makes it clearer what's going on, I've added separate warnings and tests for this.

tomaarsen:
@jaalu 
This is looking great!
I'd like to get your thoughts on one more thing: when users in in practice encounter one of these cases, then they'll encounter the following logs:
```
SetFit expects the dataset to have the columns ['label', 'text'], but only the columns ['extra_column', 'label'] were found.
  File "[sic]\setfit\issue_299.py", line 15, in <module>
    trainer.train()
  File "[sic]\setfit\src\setfit\trainer.py", line 326, in train
    self._validate_column_mapping(self.train_dataset)
  File "[sic]\setfit\src\setfit\trainer.py", line 157, in _validate_column_mapping
    raise ValueError(
ValueError: A column mapping must be provided when the dataset does not contain the following columns: {'label', 'text'}
```
If I take the role of someone who doesn't know the project well, who doesn't know about this common issue, then I would primarily look at the error message, and I may even completely miss the warning above it.
How do you feel about integrating your messages into the `ValueError`? After all, all of these warnings are directly before a `ValueError` is thrown.

- Tom Aarsen

tests/test_trainer.py
```diff
@@ -81,6 +81,23 @@ def test_trainer_raises_error_with_missing_text(self):
...
+        """If the required columns are missing from the dataset, the library should log a warning and list the columns found."""
+        dataset = Dataset.from_dict({"label": [0, 1, 2], "extra_column": ["d", "e", "f"]})
+        trainer = SetFitTrainer(
+            model=self.model, train_dataset=dataset, eval_dataset=dataset, num_iterations=self.num_iterations
+        )
+        with self.assertLogs(level=logging.WARNING) as cm:
+            try:
+                trainer._validate_column_mapping(trainer.train_dataset)
+            except ValueError:
+                pass
```
tomaarsen, line 94:
```suggestion
            with self.assertRaises(ValueError):
                trainer._validate_column_mapping(trainer.train_dataset)
```
Perhaps we can adopt `assertRaises` here?

tests/test_trainer.py
```diff
@@ -93,6 +110,46 @@ def test_column_mapping_with_missing_text(self):
...
+            train_dataset=dataset,
+            eval_dataset=dataset,
+            num_iterations=self.num_iterations,
+            column_mapping={"text_new": "text", "label_new": "label"},
+        )
+        with self.assertLogs(level=logging.WARNING) as cm:
+            try:
+                trainer._validate_column_mapping(trainer.train_dataset)
+            except ValueError:
+                pass
```
tomaarsen, line 127:
```suggestion
            with self.assertRaises(ValueError):
                trainer._validate_column_mapping(trainer.train_dataset)
```
Idem

tests/test_trainer.py
```diff
@@ -93,6 +110,46 @@ def test_column_mapping_with_missing_text(self):
...
+            test_size=0.5
+        )
+        trainer = SetFitTrainer(
+            model=self.model, train_dataset=dataset, eval_dataset=dataset, num_iterations=self.num_iterations
+        )
+        with self.assertLogs(level=logging.WARNING) as cm:
+            try:
+                trainer._validate_column_mapping(trainer.train_dataset)
+            except ValueError:
+                pass
```
tomaarsen, line 146:
```suggestion
            with self.assertRaises(ValueError):
                trainer._validate_column_mapping(trainer.train_dataset)
```
Idem

tomaarsen (commented):
I've got one small suggestion. Feel free to adopt or disregard at your discretion.

tests/test_trainer.py
jaalu: Fixed in [97232d1](https://github.com/huggingface/setfit/pull/299/commits/97232d18f587c8d87cf7d806f9efd0e285b3b0ab)

jaalu:
Good catch! Integrating the messages into the ValueErrors themselves makes sense, I think - will do that in a new commit shortly.

jaalu:
Fixed the warning from `black`.

Not to bikeshed this too heavily, but two things I'm not quite sure about:
- Should the error in the case where there is no `column_mapping` alert the user to `column_mapping` being an option?
- The case where DatasetDict is passed is really a TypeError, not a ValueError, but I'm not sure if it's worth making the distinction here (I can't imagine someone deliberately `catch`-ing for this, and tests and callers would have to catch on both TypeError and ValueError)

tomaarsen:
I think the distinction between ValueError and TypeError is only relevant in situations where someone would reasonably try to catch the error. However, your first comment is interesting. I think that would be beneficial.

Furthermore, and I recognize that this might be derailing the purpose of this PR somewhat, but I recognized the following behaviour earlier:
```python
from datasets import Dataset
from setfit import SetFitTrainer, SetFitModel

dataset = Dataset.from_dict({"sentence": ["a", "b", "c"], "label": [0, 1, 2], "extra_column": ["d", "e", "f"]})

model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")

trainer = SetFitTrainer(
    model,
    train_dataset=dataset,
    eval_dataset=dataset,
    column_mapping={"sentence": "text"}
)

trainer.train()
```
Results in:
```
ValueError: The following columns are missing from the column mapping: {'label'}. Please provide a mapping for all required columns.
```
The error is pretty clear, but having to explicitly provide a mapping from `"label"` to `"label"` is quite unintuitive. This can certainly be solved on our side by modifying the definition of `missing_columns` from `required_columns.difference(self.column_mapping.values())` to:
```
missing_columns = required_columns.difference(set(self.column_mapping.values()) | column_names)
```

That said, some quick tests show me that `_apply_column_mapping` would need to be modified as well.

jaalu:
I've added a note about `column_mapping` to the error in question.

I do think that error message could be improved, but I think it's best solved in a separate issue/PR, especially if it does affect `_apply_column_mapping`.

tomaarsen:
Agreed, great!

tomaarsen:
I don't suspect so. I don't have time to give this a full look over right now, but I'll make some time soon. Thanks for your help so far already!

tomaarsen:
I added another simple test case that covers the exact case that you wanted to solve from https://github.com/huggingface/setfit/issues/226#issuecomment-1405023578. 
Thanks a lot for this PR @jaalu, I think it'll help a lot of users out.

PR #306, opened 2023-02-08
Update run_zeroshot.py

danielkorat:
Fix bug causing `--candidate_labels` not to work.


tomaarsen (approved):
Looks good to me

lewtun (approved):
Thanks for the fix!

Issue #249, opened 2022-12-28
Sentence Pairs generation: is possible to parallelize it?


info2000:
My dataset has 20k samples, 200 labels, and 32 iterations, so that means around 128 million samples, right?
there's some way to parallelize the pairs sentences creation?
or at least to save these pairs to create one time and reuse multiple times (i.e. to train with different epochs)

Thanks

logan-markewich:
I second this. I have a dataset of 4500 classes, with 10 samples each. With 20 iterations, it looks like this will take 19 hours just for generating the pairs :(

logan-markewich:
Oh my bad. I stepped through the code, and it's actually the `sample_dataset` function that will take 19 hours 😅 Seems like something that could be improved still, as it is symptomatic of large datasets with many labels

logan-markewich:
PR is merged! This should be extremely faster now

PR #309, opened 2023-02-09
Improve dataset pre-processing speeds for large datasets

logan-markewich:
I have a dataset containing 4500 unique labels. The current implementations of sample_dataset and pair generation take more than 12 hours to complete with my dataset (I actually gave up after 12 hours). This was reported in https://github.com/huggingface/setfit/issues/249

The new implementation allows the dataset sampling and pair generation to finish in minutes.

In the future, I would like to improve the multilabel and cossim pair generation, but it's not as straightforward. The best advice I can give is that np.where can be pretty slow, so avoiding it helps a lot.

Let me know if you want anything changed, I am happy to make edits!

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -350,7 +350,7 @@ class SetFitTrainer:
             else:
                 train_examples = []
 
-                for _ in trange(self.num_iterations, desc="Generating Training Pairs"):
+                for _ in trange(self.num_iterations, desc="Generating Training Pairs", disable=not show_progress_bar):
                     if self.model.multi_target_strategy is not None:
                         train_examples = sentence_pairs_generation_multilabel(
                             np.array(x_train), np.array(y_train), train_examples


logan-markewich:
I'll fix those checks sometime today 😅

logan-markewich:
I think this is ready to merge! 🚢  (or at least a review 👍🏻)

tomaarsen:
Hello @logan-markewich!

Thanks for these changes. I ran this PR locally last night to see if the change in sampling resulted in any change in performance, and it seems that it doesn't (which is good!). Beyond that, I'm interested in getting a better understanding of the time efficiency before and after this PR. I may have some time for this at the start of next week, and otherwise perhaps you can look into this, too.
That way, we both have a better understanding of exactly how useful this PR is 🎉 

I have some small nitpicks from my testing last night, I'll write those out in a quick review.

- Tom Aarsen

src/setfit/data.py
```diff
@@ -160,14 +165,15 @@ def create_samples(df: pd.DataFrame, sample_size: int, seed: int) -> pd.DataFram
...
-    all_samples = concatenate_datasets(samples)
+
+    df = shuffled_dataset.to_pandas()
+    df = df.groupby(label_column)
+
+    # sample num_samples, or at least as much as possible
+    df = df.apply(lambda x: x.sample(min(num_samples, len(x))))
+    df = df.reset_index(drop=True)
+
+    all_samples = Dataset.from_pandas(df)
```
tomaarsen, line 176: I love the removal of all loops 👍

src/setfit/modeling.py
```diff
@@ -677,19 +677,21 @@ def sentence_pairs_generation(sentences, labels, pairs):
     # labels to indicate if a pair is positive or negative
 
     num_classes = np.unique(labels)
-    idx = [np.where(labels == i)[0] for i in num_classes]
+    label_to_idx = {x: i for i, x in enumerate(num_classes)}
+    positive_idxs = [np.where(labels == i)[0] for i in num_classes]
+    negative_idxs = [np.where(labels != i)[0] for i in num_classes]
 
-    for first_idx in range(len(sentences)):
+    for first_idx in tqdm(range(len(sentences)), desc="Gen Pairs"):
```
tomaarsen, line 684: I have a few comments here:
1. Can we use [`trange(...)`](https://tqdm.github.io/docs/tqdm/#trange) instead of `tqdm(range(...))`?
2. Perhaps `Generating Training Pairs` or `Generating Contrastive Training Pairs` is a bit clearer than `Gen Pairs`?
3. Can we use [`leave=False`](https://github.com/tqdm/tqdm#parameters)? Otherwise we end up with e.g. 20 completed `tqdm` bars after the pairs are generated.

src/setfit/modeling.py
```diff
@@ -699,7 +701,7 @@ def sentence_pairs_generation(sentences, labels, pairs):
 def sentence_pairs_generation_multilabel(sentences, labels, pairs):
     # Initialize two empty lists to hold the (sentence, sentence) pairs and
     # labels to indicate if a pair is positive or negative
-    for first_idx in range(len(sentences)):
+    for first_idx in tqdm(range(len(sentences)), desc="Gen Pairs"):
```
tomaarsen, line 704: Idem

src/setfit/modeling.py
```diff
@@ -728,7 +730,7 @@ def sentence_pairs_generation_cos_sim(sentences, pairs, cos_sim_matrix):
 
     idx = list(range(len(sentences)))
 
-    for first_idx in range(len(sentences)):
+    for first_idx in tqdm(range(len(sentences)), desc="Gen Pairs"):
```
tomaarsen, line 733: Idem.

src/setfit/modeling.py
logan-markewich: I totally agree! And actually, it's a little weird/confusing to show the progress for each iteration. 

Instead, I moved the `trange` into `trainer.py`, so the user just sees the top-level progress for all iterations being generated

logan-markewich:
Besides those changes, I'll generate some graphs later today to show the difference in speed for the two functions I improved 👍🏻

logan-markewich:
I setup some dummy data like below, and just adjusted the `num_train_labels` variable to measure performance.

```
        num_train_labels = 256
        data_size = 1000
        train = {'text': [], 'label': []}

        for label in range(num_train_labels):
            for feat in range(data_size):
                train['text'].append('test')
                train['label'].append(label)

        train = Dataset.from_dict(train)

        from setfit.modeling import sentence_pairs_generation
        import time
        sample_start = time.time()
        train = sample_dataset(train, label_column='label', num_samples=8)
        sample_end = time.time()

        gen_start = time.time()
        sentence_pairs_generation(train['text'], train['label'], [])
        gen_end = time.time()

        print('Sample time: ', sample_end - sample_start)
        print('Generating pairs time: ', gen_end - gen_start)
```

Looks like generate_sentence_pairs performance is about the same

sample_dataset is extremely faster though!

image

src/setfit/modeling.py
```diff
@@ -21,7 +21,7 @@
 from sklearn.multiclass import OneVsRestClassifier
 from sklearn.multioutput import ClassifierChain, MultiOutputClassifier
 from torch.utils.data import DataLoader
-from tqdm.auto import tqdm
+from tqdm import trange
```
tomaarsen, line 24:
```suggestion
from tqdm.auto import trange
```
The `tqdm.auto` import has support for progress bars in notebooks, whereas simply `from tqdm import ...` does not. See the docs [here](https://tqdm.github.io/docs/shortcuts/). (Although I suppose `tqdm.autonotebook` would also work as we don't use asyncio)

src/setfit/modeling.py
```diff
@@ -297,7 +297,7 @@ def fit(
             criterion = self.model_head.get_loss_fn()
             optimizer = self._prepare_optimizer(learning_rate, body_learning_rate, l2_weight)
             scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=5, gamma=0.5)
-            for epoch_idx in tqdm(range(num_epochs), desc="Epoch", disable=not show_progress_bar):
+            for epoch_idx in trange(num_epochs, desc="Epoch", disable=not show_progress_bar):
```
tomaarsen, line 300: Good call!

src/setfit/trainer.py
```diff
@@ -7,6 +7,7 @@
 from sentence_transformers.datasets import SentenceLabelDataset
 from sentence_transformers.losses.BatchHardTripletLoss import BatchHardTripletLossDistanceFunction
 from torch.utils.data import DataLoader
+from tqdm import trange
```
tomaarsen, line 10:
```suggestion
from tqdm.auto import trange
```
As before. (`tqdm.autonotebook` is also fine)

src/setfit/trainer.py
```diff
@@ -349,7 +350,7 @@ def train(
             else:
                 train_examples = []
 
-                for _ in range(self.num_iterations):
+                for _ in trange(self.num_iterations, desc="Generating Training Pairs"):
```
tomaarsen, line 353:
```suggestion
                for _ in trange(self.num_iterations, desc="Generating Training Pairs", disable=not show_progress_bar):
```

tomaarsen:
Now those are some very impressive results. Quite happy to see such enormous improvements! 
I played around with this briefly locally and it all seems to work as intended. All that remains now is some nitpicks :)

logan-markewich:
Nitpicks settled, thanks for the suggestions 💪🏻

PR #321, opened 2023-02-28
Fix output of hyperparameter search notebook

EdAbati:
Hi :)

While going through the examples, I have noticed that one of the cells of the `text-classification_hyperparameter-search.ipynb` notebook raised an Exception.

I just rerun that notebook, and now it looks fine.


tomaarsen:
So this is only reran, no other changes? The diff, and even ReviewNB, is a bit of a mess as you can imagine, haha.
It looks good at a glance.

EdAbati:
Yes, just rerun it in Colab no changes to the code/markdown!

Oh yeah, the diff in ReviewNB looks messy too. Mmm… The diff on VSCode looked good. 😕 I actually made sure to only commit the outputs to keep the diff small (because Colab changed the metadata of each cell a bit too).

Are there any tricks to make ReviewNB show a more helpful diff?

tomaarsen:
I'm guessing it's not possible. It's a bit surprising, really, because it should just be able to recognize which cells correspond between main and this PR. After all, the number of cells didn't change, and only the outputs differ...

That said, this is great! Thank you for taking the time to fix this for us! I have no issues merging this.

PR #327, opened 2023-03-07
[feat] cpu-load support for differentiable head

karter-liner:
If we train a differentiable head on a CUDA machine, the head checkpoint cannot be loaded via Joblib on CPU machine (Serialization error).
To address this issue, I modified the code to store the head checkpoint with CPU mapping.
When we load the head checkpoint, we map it to the target device.


tomaarsen:
Hello!

Apologies for the late response. This PR is looking great. We've been working on a v1.0.0 release with a large refactor, so I've taken your changes and applied them on top of that branch in #444. I've marked you as a co-author on the PR. 

I'll be closing this one in favor of that one. Hope that's okay!

- Tom Aarsen

Issue #329, opened 2023-03-10
SetFit trainer.evaluate() TypeError


splevine:
![image](https://user-images.githubusercontent.com/4715331/224421976-5dc8c81c-41f1-40a5-9bdc-f9d12659870f.png)

In colab I'm getting this error when trying to use built in evaluate for setfit 0.6.0

I'm assuming it has to do with the change for `predict()` where I add in parameter `as_numpy=True`, but that's not an option for `evaluate()`

pastorsj-dlt:
This happens for me as well. Hoping we can get some insight on this.

pastorsj-dlt:
Actually, I found that I may need to encode the labels as integers rather than leave them as strings. I don't know whether that fixes the issue, but perhaps that is the problem.

tomaarsen:
Hello @splevine and @pastorsj-dlt, thank you for bringing this to my attention. This is indeed an issue. #331 should improve the situation, but it will not fix it. This is because the [evaluate](https://github.com/huggingface/evaluate) library that we use does not support string labels in its `accuracy` metric. This can be counteracted by supplying a different `metric`, e.g. a function, that computes some metric **with** support of strings. 
There's some info regarding the callable in the `SetFitTrainer` docstring:
https://github.com/huggingface/setfit/blob/0420165a14431809a63ffc36f2342b6ed8190d0c/src/setfit/trainer.py#L43-L45

- Tom Aarsen

tomaarsen:
#439 should fully solve this, as the Trainer now uses a [LabelEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html) to turn the labels into integers if the labels are strings.

PR #331, opened 2023-03-13
Prevent TypeError on `model.predict` when using string labels

tomaarsen:
Hello!

## Pull Request overview
* Prevent TypeError on `model.predict` when using string labels.
* Added a test case to show correct behaviour.

## Details
When training with string labels (which is not strictly recommended, but possible), then `model.predict` broke as of the latest version. See the following script to reproduce:

### Reproduction
```python
from datasets import Dataset, load_dataset

dataset = Dataset.from_dict(
    {"text": ["positive sentence", "negative sentence"], "label": ["positive", "negative"]}
)
model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2")
trainer = SetFitTrainer(
    model=model,
    train_dataset=dataset,
    eval_dataset=dataset,
    num_iterations=1,
)
trainer.train()
# This used to fail due to "TypeError: can't convert np.ndarray of type numpy.str_.
# The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool."
model.predict(["another positive sentence"])
```

This resulted in 
```
  File "[sic]demo_string_issue.py", line 17, in <module>
    model.predict(["another positive sentence"])
  File "[sic]src\setfit\modeling.py", line 419, in predict
    outputs = torch.from_numpy(outputs)
TypeError: can't convert np.ndarray of type numpy.str_. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.
```
See also #329, which shows this same issue, but for `evaluate` (which calls `predict` behind the scenes).

### Why do we get this error?
Consider the following lines in the `predict` method:
https://github.com/huggingface/setfit/blob/0420165a14431809a63ffc36f2342b6ed8190d0c/src/setfit/modeling.py#L414-L421
And consider the scenario with the (default) non-differentiable head and `as_numpy=False`. In this case, we reach line 419 and call `torch.from_numpy`. However, `outputs` has dtype `<U8`, where the `U` indicates that the type is a unicode string. There is no Torch tensor equivalent of this type, and thus we get the error shown above.

### The fix
The fix is simply to prevent calling `torch.from_numpy` if the head outputs a numpy array with strings.

## Note
The issue from #329 isn't exactly fixed, calling `evaluate` using string labels still fails, as the [evaluate](https://github.com/huggingface/evaluate) library does not support string labels in its `accuracy` metric. This can be counteracted by supplying a different `metric`, e.g. a function, that computes some metric **with** support of strings.

- Tom Aarsen

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -481,10 +481,11 @@ class SetFitModel(PyTorchModelHubMixin):
     def __call__(self, inputs):
         return self.predict(inputs)
 
-    def _save_pretrained(self, save_directory: str) -> None:
+    def _save_pretrained(self, save_directory: Union[Path, str]) -> None:
+        save_directory = str(save_directory)
         self.model_body.save(path=save_directory, create_model_card=False)
         self.create_model_card(path=save_directory, model_name=save_directory)
-        joblib.dump(self.model_head, f"{save_directory}/{MODEL_HEAD_NAME}")
+        joblib.dump(self.model_head, str(Path(save_directory) / MODEL_HEAD_NAME))
 
     @classmethod
     def _from_pretrained(


tomaarsen:
Test failures are unrelated, solved by #332.

PR #341, opened 2023-03-31
Update modeling.py

kobiche:
I propose the following change in the _from_pretrained() method from SetFitModel:

- Call init method from cls and not SetFitModel, otherwise overriding methods from this class is not possible.


tomaarsen (approved):
Looks good! Thanks for this.

PR #343, opened 2023-04-03
Allow distillation training with models using differentiable heads

### LINKED ISSUE #342: Error in DistillationTrainer: not valid argument type for student_model call
I tried using the Distillation method for training a model with a  diff. classification head i.e. SetFitHead. However, there seems to be a bug while calling the student model in the trainer_distillation.py.
The error is the following:
`TypeError: linear(): argument 'input' (position 1) must be Tensor, not numpy.ndarray`

It comes in the line 
```python
 # **************** student training ****************
x_train_embd_student = self.teacher_model.model_body.encode(x_train)
y_train = self.teacher_model.model_head.predict(x_train_embd_student)
``` 
I believe `x_train_embd_student` should be transformed to torch.Tensor().

tomaarsen:
Closes #342

Hello!

## Pull Request overview
* Allow distillation training with differentiable models
  * Both distilling from a differentiable model into any other model, and
  * distilling from a logistic regression model into a differentiable head.
* Add tests for this.

## Details
See #342 for details. The encode and predictions in the distillation trainer were too naive. They lacked conversions between numpy arrays and torch tensors if there are discrepancies between the types of heads between the teacher and student models. Furthermore, the differentiable head teacher model fails even earlier on due to the `encode` automatically returning a numpy array, which was then used with the differentiable head.

I've made the required type transformations, alongside a testing function that tries all 4 combinations of teachers and students.

- Tom Aarsen

### DIFF
--- a/src/setfit/trainer_distillation.py
+++ b/src/setfit/trainer_distillation.py
@@ -192,8 +192,14 @@ class DistillationSetFitTrainer(SetFitTrainer):
                 train_examples = []
 
                 # **************** student training ****************
-                x_train_embd_student = self.teacher_model.model_body.encode(x_train)
+                x_train_embd_student = self.teacher_model.model_body.encode(
+                    x_train, convert_to_tensor=self.teacher_model.has_differentiable_head
+                )
                 y_train = self.teacher_model.model_head.predict(x_train_embd_student)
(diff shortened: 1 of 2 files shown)


tomaarsen:
Sure! I think this is ready regardless :) Feel free to let me know if you find any issues with it.
You can install it with
```
pip install git+https://www.github.com/huggingface/setfit.git
```

Issue #110, opened 2022-10-18
more metrics addition (i.e f1score, precision ) in the trainer.evaluate()


snayan06:
was just checking the code and saw only accuracy as a metrics, are we planning to add more metrics?

lewtun:
Hi @snayan06 thanks for your interest in `setfit`! Although `accuracy` is the default metric, you can specify a different one when you create the `SetFitTrainer`, e.g.:

```
from setfit import SetFitTrainer

trainer = SetFitTrainer(model=model, train_dataset=train_dataset, metric="f1")
```

Currently, we support any classification metric from the `evaluate` library: https://huggingface.co/docs/evaluate/index

snayan06:
ok thanks for answering , was just looking to contribute and thought that was not there in setfit, and would have been helpful to add. thanks to pointing to the the library as well.

snayan06:
Is there any way u guys track features and people can contribute to build these features, or it is just in the issues , new to open source but have been using huggingface/sbert models from a long time and thought if i can contribute in some way that would be good.

kamilc-bst:
What if I have multi-label classification problem + I want to evaluate my eval dataset via "f1" + specific average strategy like 'weighted' or 'micro' ? Is it supported ?

nsorros:
> Currently, we support any classification metric from the `evaluate` library: https://huggingface.co/docs/evaluate/index

@lewtun one limitation though seems to be that in case of non binary labels there is no way to pass an average strategy. this would need to happen here
https://github.com/huggingface/setfit/blob/a06be0efae1f71357aa357af8df42175f6adc91e/src/setfit/trainer.py#L303

So in those cases you need to fallback to accuracy or write your own trainer. Would it be in scope to support those evaluations? I am happy to open a PR for this.

jmwoloso:
shouldn't this line in the same file (and method) handle multi-label though?

https://github.com/huggingface/setfit/blob/a06be0efae1f71357aa357af8df42175f6adc91e/src/setfit/trainer.py#L295

jmwoloso:
assuming I'm not missing something major, this seems like an easy fix. here's my current implementation that I'm testing. happy to start a formal PR for this if desired @lewtun 

https://github.com/huggingface/setfit/compare/main...jmwoloso:setfit:fix_multilabel_metrics?expand=1

PR #353, opened 2023-04-13
Add metric keyword arguments, e.g. add "average" strategy to f1

### LINKED ISSUE #345: Training a ternary Classifier: "Average" setting for Metrics "precision", "recall" and "f1"
Hi,
I trained a **ternary classifier** using SetFit and wanted evaluate the metrics **"accuracy", "precision", "recall" and "f1"**. However, I got the following error for all of these metrics except "accuracy":

        ValueError: Target is multiclass but average='binary'. Please choose another average setting, one of [None, 'micro', 'macro', 'weighted'].

To be able to use **other average settings** I overwrote the "**SetFitTrainer**" class with "**MySetFitTrainer**" as follows, overwriting the "evaluate" method (most of it is copy & paste, except for the two "if-elsed" calls to "metric_fn.compute(...)"):

        logger = logging.get_logger(__name__)

        class MySetFitTrainer(SetFitTrainer):
            def __init__(self, **kwargs):
                super(MySetFitTrainer, self).__init__(**kwargs)
            
            def evaluate(self, average=None, **kwargs):
                self._validate_column_mapping(self.eval_dataset)
                eval_dataset = self.eval_dataset
            
                if self.column_mapping is not None:
                    logger.info("Applying column mapping to evaluation dataset")
                    eval_dataset = self._apply_column_mapping(self.eval_dataset, self.column_mapping)
            
                x_test = eval_dataset["text"]
                y_test = eval_dataset["label"]
            
                logger.info("***** Running evaluation *****")
                y_pred = self.model.predict(x_test)
            
                if isinstance(self.metric, str):
                    metric_config = "multilabel" if self.model.multi_target_strategy is not None else None
                    metric_fn = evaluate.load(self.metric, config_name=metric_config)
                    print("eval:", self.metric, average if average is not None else None)
                    if (self.metric == "accuracy"):
                        result = metric_fn.compute(predictions=y_pred, references=y_test)
                    else:
                        result = metric_fn.compute(predictions=y_pred, references=y_test, average=average)
                    print("eval:", result)
                    return result
            
                elif callable(self.metric):
                    return self.metric(y_pred, y_test)
            
                else:
                    raise ValueError("metric must be a string or a callable")
This works as expected and when passing in different values for  the average parameter the result is the following:
grafik
**My question is: For average "micro" in row 1 all metrics give the identical value (i.e. 0.724138). Is this a bug or expected behavior?** This is unclear to me. Help is appreciated. Thanks and best regards,
Stefan

tomaarsen:
Closes #345, #110

Hello!

## Pull Request overview
* Add `metric_kwargs` parameter to the `SetFitTrainer`. These parameters are fed to the `evaluate` metric.
* Add test to show that you can now use "f1" with a multi-label model, which was frequently requested.

## Details
The following snippet used to fail:
```python
dataset = Dataset.from_dict({"text": ["", "a", "b", "ab"], "label": [[0, 0], [1, 0], [0, 1], [1, 1]]})
    "sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="one-vs-rest"
)

trainer = SetFitTrainer(
    model=model,
    train_dataset=dataset,
    eval_dataset=dataset,
    metric="f1",
    num_iterations=5,
)

trainer.train()
metrics = trainer.evaluate()
```
With:
```
    ValueError: Target is multiclass but average='binary'. Please choose another average setting, one of [None, 'micro', 'macro', 'weighted'].
```

## The solution
Users have suggested to add `metric_average` to tackle this (e.g. https://github.com/huggingface/setfit/issues/110#issuecomment-1335852699), but I think it's best if we simply allow users to pass a dictionary of keyword arguments. Note that these keyword arguments are only used if the metric is an `evaluate` metric, not if it's a callable.

I've also added a test to show that this allows for "f1" on multi-label models.

I'd like to include this in 0.7.0.

- Tom Aarsen

### DIFF
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -468,7 +468,9 @@ def test_trainer_evaluate_with_strings():
 
 def test_trainer_evaluate_multilabel_f1():
     dataset = Dataset.from_dict({"text_new": ["", "a", "b", "ab"], "label_new": [[0, 0], [1, 0], [0, 1], [1, 1]]})
-    model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="one-vs-rest")
+    model = SetFitModel.from_pretrained(
+        "sentence-transformers/paraphrase-albert-small-v2", multi_target_strategy="one-vs-rest"
+    )
 
     trainer = SetFitTrainer(
         model=model,


lewtun (approved):
Thanks for the fix @tomaarsen ! I agree using `kwargs` is the most flexible way to handle these options - for v1 of `setfit` we might want to consider having a dedicated `compute_metrics` method to enable full flexibility (like in `transformers` [here](https://huggingface.co/docs/transformers/v4.28.0/en/main_classes/trainer#transformers.Trainer.compute_metrics)), but that's a separate discussion :)

tomaarsen:
Don't we already have that? Users can also specify a callable. The only difference is that the Transformers version gets an `EvalPrediction` instance, while our version gets `y_pred` and `y_test`. Or do you mean using a `compute_metrics` method as the default?

Issue #91, opened 2022-10-07
Using Setfit for similarity classification


castafra:
Hello,
I would like to test this promising framework on a similarity classification task. So basically, I have got a dataset with 3 columns: (sentence1,sentence2,label). From what I understand, currently it is only possible to train on a single sentence classification problem.
Is there a get around to use Setfit for a pair sentence classification problem ? If not, would it be possible to add this feature in a future integration ? 

Thank you in advance

lewtun:
Hi @castafra thanks for your interest in our work! I know @orenpereg ran some experiments on 2-sentence tasks like GLUE, but it wasn't entirely clear how one should create the text/label triples one needs for the contrastive learning step.

Perhaps he can share some more details on what worked / didn't work :)

castafra:
Hi @orenpereg , could you share any feedbacks on your experiments ? 
Thanks

orenpereg:
Hi @castafra, Thanks for your question. Currently Setfit does not support two sentences similarity classification tasks and it's not trivial to apply SetFit for that. It will probably require changes within the Sentence Transformer itself. We do have plans to extend SetFit and one of the options is to examine similarity classification task. Having said that, i did play a bit with it, as a workaround, I simply concatenated sentence1 and sentence2 and input pairs of 2 concatenated sentences to the contrastive training process. The results weren't so good. SetFit was on-par with the baseline model (a 'standard' BERT cross encoder) when using few-shot data (8,16,32 samples). The bottom line is that this workaround is not recommended.

jpzhangvincent:
I also work on the similarity classification task. I think it would be great to extend the framework for that task. 
It seems `setFit` uses the underlying `sentence_transformer` library, right? FWIW,  the example shows how to use `sentence_transformer` for similarity learning - https://www.sbert.net/examples/training/sts/README.html . I'm wondering whether we just need to implement a `sklearn-learn` classifier model to take two fine-tuned embeddings as inputs to predict the label.

rjurney:
[...]

PR #354, opened 2023-04-13
Add support for sentence pair classification

juanjucm:
Hi!
This PR solves #91

### PR Description

This PR focus on enhancing setfit for supporting sentence pair classification task. According to [Sentence Transformers docs](https://www.sbert.net/examples/training/sts/README.html) (also mentioned in #91), I think it could be possible to add this functionality.

### PR Context

A common task for BERT-like models is **sentence pair classification**.  Not only for semantic similarity, as suggested in the original discussion, but for other tasks too. This is, passing the model sentences like [CLS]Sentence1[SEP]Sentence2[SEP].

I understand that this change affects a big part of the functionality / code, so I'm creating this PR just as draft and as a starting point of discussion. I would like to know about the viability of this addition and also extra ideas for doing this right.

In the meantine, I will keep working on this and updating this PR.

### DIFF
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -234,8 +234,9 @@ class SetFitDataset(TorchDataset):
     A dataset for training the differentiable head on text classification.
 
     Args:
-        x (`List[str]`):
-            A list of input data as texts that will be fed into `SetFitModel`.
+        x (`Union[List[str], List[List[str]]]`):
+            A list of input data as texts that will be fed into `SetFitModel`. 
+            If x is a list of lists, we will be creating a dataset for sentence pair classification task. 
         y (`Union[List[int], List[List[int]]]`):
             A list of input data's labels. Can be a nested list for multi-label classification.
         tokenizer (`PreTrainedTokenizerBase`):
@@ -247,7 +248,7 @@ class SetFitDataset(TorchDataset):
 
     def __init__(
         self,
-        x: List[str],
+        x: Union[List[str], List[List[str]]],
         y: Union[List[int], List[List[int]]],
         tokenizer: "PreTrainedTokenizerBase",
         max_length: int = 32,
@@ -263,14 +264,26 @@ class SetFitDataset(TorchDataset):
         return len(self.x)
 
     def __getitem__(self, idx: int) -> Tuple[TokenizerOutput, Union[int, List[int]]]:
-        feature = self.tokenizer(
-            self.x[idx],
-            max_length=self.max_length,
-            padding="max_length",
-            truncation=True,
-            return_attention_mask="attention_mask" in self.tokenizer.model_input_names,
-            return_token_type_ids="token_type_ids" in self.tokenizer.model_input_names,
-        )
+        # Sentence pair classification
+        if isinstance(self.x[0], list) and len(self.x[0] == 2):
+            feature = self.tokenizer(
+                self.x[0][idx],
+                self.x[1][idx],
+                max_length=self.max_length,
+                padding="max_length",
+                truncation=True,
+                return_attention_mask="attention_mask" in self.tokenizer.model_input_names,
+                return_token_type_ids="token_type_ids" in self.tokenizer.model_input_names,
+            )
+        else:
+            feature = self.tokenizer(
+                self.x[idx],
+                max_length=self.max_length,
+                padding="max_length",
+                truncation=True,
+                return_attention_mask="attention_mask" in self.tokenizer.model_input_names,
+                return_token_type_ids="token_type_ids" in self.tokenizer.model_input_names,
+            )
         label = self.y[idx]
 
         return feature, label
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -278,7 +278,7 @@ class SetFitModel(PyTorchModelHubMixin):
 
     def fit(
         self,
-        x_train: List[str],
+        x_train: Union[List[str], List[List[str]]],
         y_train: Union[List[int], List[List[int]]],
         num_epochs: int,
         batch_size: Optional[int] = None,
@@ -323,7 +323,7 @@ class SetFitModel(PyTorchModelHubMixin):
 
     def _prepare_dataloader(
         self,
-        x_train: List[str],
+        x_train: Union[List[str], List[List[str]]],
         y_train: Union[List[int], List[List[int]]],
         batch_size: Optional[int] = None,
         max_length: Optional[int] = None,
@@ -711,6 +711,7 @@ def sentence_pairs_generation(sentences, labels, pairs):
 def sentence_pairs_generation_multilabel(sentences, labels, pairs):
     # Initialize two empty lists to hold the (sentence, sentence) pairs and
     # labels to indicate if a pair is positive or negative
+
     for first_idx in range(len(sentences)):
         current_sentence = sentences[first_idx]
         sample_labels = np.where(labels[first_idx, :] == 1)[0]


tomaarsen:
Hello @juanjucm,

This is a good start to sentence pair classification, I think. However, in order for us to justify the maintenance cost of this, could you provide me a use case for this form of classification? (or multiple) If there is indeed a reasonable use case, and the setfit approach works well, then I think we could adopt this task.

Now, I also want to point out that my intention is to merge #265 relatively soon, pushing us into v1.0.0 with a large refactor. This may complicate things, as PRs made will be automatically outdated and require a decent bit of changes. I want to warn you of this.

From your description, is my assumption correct that you're trying to structure this implementation primarily as a data preprocessing step, i.e. converting the sentence pair classification problem into a more traditional text classification problem by creating `[CLS]Sentence1[SEP]Sentence2[SEP]`? However, this seems to match Oren Pereg's work as described here: https://github.com/huggingface/setfit/issues/91#issuecomment-1285344061.
He mentions that this simple approach does not outperform finetuning BERT. 

If the alternative solution requires changes on the Sentence Transformer level, then I think it might be best suited for a separate repository, i.e. not in `setfit`. This essentially involves carrying out research to try and adapt setfit to sentence pair classification, and then developing a Python package for it.

I'd love to hear your thoughts on all of this,

- Tom Aarsen

juanjucm:
Hi @tomaarsen, thanks for your feedback!

> This is a good start to sentence pair classification, I think. However, in order for us to justify the maintenance cost of this, could you provide me a use case for this form of classification? (or multiple) If there is indeed a reasonable use case, and the setfit approach works well, then I think we could adopt this task.

Sentence pair classification is useful for all tasks involving two (or even more) texts input features. 
- Maybe the most common use cases would be sentence similarity or sentence entailment, where you want to classify the degree of relation between two sentences. Check [this post](https://towardsdatascience.com/fine-tuning-pre-trained-transformer-models-for-sentence-entailment-d87caf9ec9db) as an example of this kind of tasks.
- Of course, this could be extended if working in a multilingual scenario.  For example, we may want to assess how similar an english and a translated text is by training a model with [en, multi-lang] pairs.
- Apart from these tasks, I've also encounter cases in which adding an extra text feature to the input could result on a model taking more contextualized / informed decissions.

Having this kind of tasks supported by `setfit` could be amazing for low resource scenarios where labeled data is not easy to find.

> From your description, is my assumption correct that you're trying to structure this implementation primarily as a data preprocessing step, i.e. converting the sentence pair classification problem into a more traditional text classification problem by creating [CLS]Sentence1[SEP]Sentence2[SEP]? However, this seems to match Oren Pereg's work as described here: https://github.com/huggingface/setfit/issues/91#issuecomment-1285344061.
He mentions that this simple approach does not outperform finetuning BERT.

Indeed, this would imply a modification on the preprocessing step. HuggingFace tokenizers for BERT-like models allow to input a pair of sentences, generating a tokenization that looks as follows `[CLS]Sentence1[SEP]Sentence2[SEP]` . The `tokenizer` generates an extra [SEP] token encoded in `token_ids`. It also returns a specific `token_type_ids` in which both sentences are given 1 and 0 values respectively. Here you have a [great post](https://huggingface.co/docs/transformers/glossary#token-type-ids) on how this tokenization process is done.

Regarding Oren Pereg's work, is not really the same. If I am not mistaken, what he did was a workaround for actual sentece-pair classification. He tried directly concatenating two sentences and inputing that into the tokenizer and then into the model. This resulting on the model interpreting the input as a single sentence. 
In a sentence-pair setup, after the proper tokenization, the model will understand (by looking at  `token_types` and `token_type_ids`) that there are multiple sentence, hence taking that information when generating the embeddings and extracting semantic relationships.

> f the alternative solution requires changes on the Sentence Transformer level, then I think it might be best suited for a separate repository, i.e. not in setfit. This essentially involves carrying out research to try and adapt setfit to sentence pair classification, and then developing a Python package for it.

These last week I've been taking a look at `sentence-transformers` and `setfit`code and I think it is mostly feasible. The only point of conflict I'm encountering is in these functions: `sentence_pairs_generation_multilabel()`, `sentence_pairs_generation_cos_sim()`. This is because of SentenceTransformer's `InputExample()` structure not allowing [sentence-pair], [sentence-pair] as input in the `text` argument.
It might be a workaround so the tokenizer can be aplied first and then create the contrastive pairs with the encodings or something like that. Another solution could be to use raw HuggingFace, but in this case, this would probably fit better as a separate project as you mentioned.

Sorry for this big text 😅 I think this is super exciting and I'm really loking forward to collaborate on this, so count on me for it if needed! 

Juanju

tomaarsen:
Juanlu, thank you for your wonderful response! The potential tasks are very clear, and I agree that this addition could now be worth the maintenance cost. I'm familiar with `token_type_ids`, although I'm not certain whether all `SentenceTransformer` models support `token_type_ids` as inputs. That should be a minor issue though, as I'm confident that you can load SentenceTransformer models that do support it.

If Oren's approach was indeed only sentence concatenation, then I understand that this may not work as well as also explicitly providing the `token_type_ids`: the attention would be totally different between the two approaches.

I'm glad to hear that you've delved into the SentenceTransformers code as well. This has historically been a bit of a bottleneck for us, as we couldn't modify how their [`fit` method](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L575) works. However, I'm currently working on an unrelated idea for SetFit that would benefit greatly from overriding their `fit` method, so I think this PR may also be able to make some changes there if necessary. For example, we could stop having to use their `InputExample` instances if we override their [`smart_batching_collate` method](https://github.com/UKPLab/sentence-transformers/blob/3e1929fddef16df94f8bc6e3b10598a98f46e62d/sentence_transformers/SentenceTransformer.py#L529).

I'm excited for us to try this use case. Two final notes:
* My hope is to merge #265 asap, as soon as my co-maintainer has time to review it. It's a relatively large refactor, which would likely result in some nasty merge conflicts on this PR.
* We may benefit from creating an initial, hacky implementation to try and see if the approach indeed works well or not. No use in writing beautiful, well integrated code if the model isn't suited for the task and it's not worth doing at all.

- Tom Aarsen

juanjucm:
Hi @tomaarsen, thanks for your thoughts. Happy to hear this addition makes sense! 

Following your incoming merge for #265 , if you agree, I will close this PR and open another one once the new version is out.

In the meantime, I will try to research how to do the sentence pair addition with current `sentence transformers `and `setfit` implementation.

Juanju

tomaarsen:
Perfect! Please let me know whether the approach works well! Perhaps a good baseline is merely finetuning RoBERTa-base or XLM-RoBERTa-base on (subsections of) a relatively small dataset, much like the [SetFit paper](https://arxiv.org/pdf/2209.11055.pdf). At least, I assume this is a reasonably commonly used model for the pair classification task.

- Tom Aarsen

ramonziai:
Any chance of reviving this? I'm looking at SetFit for short answer grading (which often involves comparisons to reference answers) and this feature would allow few-shot cross-prompt models, instead of one model for every prompt.

ramonziai:
@juanjucm I'm very interested in this feature. Did you get a chance to follow up on it?

FahadEbrahim:
This will be interesting if implemented.

rjurney:
@juanjucm FWIW, I would be very happy if you made this work!

haukelicht:
@juanjucm, have you had a look at the [`liqfit`](https://github.com/Knowledgator/LiqFit) library yet? They promise efficient few-shot fine-tuning of cross-encoders which generally focus on learning to predict labels/scores at the level of pairs of text. I haven't had the time to look at all the modules this library provides, but they explain [in their documentation](https://docs.knowledgator.com/docs/frameworks/liqfit/quick-start#using-classificationhead) that they provide a classification head sutiabe for the sentence pair classification task. I wonder whether this classification head could be combined with a sentence transformer body in the `setfit` framework?

juanjucm:
Thanks a lot for sharing @haukelicht! I didn't know that library. 

I didn't have time to test it yet nor inspecting the code. However, the challenge with adapting setfit is not related with the classification head, but with the encoder's input(SentenceTransformers `InputExample` class). At the end, the classification head is agnostic to whether the input consists on one or N strings (it's just feeded with whatever the final encoder's pooled embeding(s) is).

As I said, I didn't check the code yet, but I guess liqfit works for this use case because it relies on the base Transformer classes. I will take a look at it!

Issue #339, opened 2023-03-28
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.


lenhhoxung86:
I used the following model init:

```
    model = SetFitModel.from_pretrained(
        "sentence-transformers/paraphrase-mpnet-base-v2",
        multi_target_strategy="one-vs-rest",
        use_differentiable_head=True,
        head_params={"out_features": len(categories)},
    )
```

And it threw the following error in the evaluation step (training was fine):

```
***** Running evaluation *****
Traceback (most recent call last):
  File "fewshot_classify_notes_multilabel_kfold.py", line 113, in <module>
    metrics = trainer.evaluate()
  File "/home/ubuntu/anaconda3/envs/pytorch_p37/lib/python3.7/site-packages/setfit/trainer.py", line 424, in evaluate
    return self.metric(y_pred, y_test)
  File "fewshot_classify_notes_multilabel_kfold.py", line 68, in compute_metrics
    predictions = predictions.numpy()
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

If the model is initialized as follows:

```
    model = SetFitModel.from_pretrained(
        "sentence-transformers/paraphrase-mpnet-base-v2",
        multi_target_strategy="one-vs-rest"
    )
```

it worked without a problem.

How do I solve it?

tomaarsen:
In your custom `compute_metrics` method, line 68 of `fewshot_classify_notes_multilabel_kfold.py`, the `predictions` tensor is still on the GPU, meaning that you must first move it to the CPU before you can convert it to numpy (as numpy isn't possible on the GPU).

As shown in the TypeError, you must first use `predictions.cpu().numpy()`. 
There is a chance that you'll encounter a similar error afterwards, perhaps complaining about "detaching". This refers to how the `predictions` tensor may still be "attached" to the computational gradient graph. If so, you can use `predictions.detach().cpu().numpy()`.

I hope that helps!

- Tom Aarsen

lenhhoxung86:
It works, thanks.
So in two settings, the first time the `predictions` is on cpu and the second time the `predictions` is on gpu.
I would suggest to put `predictions` always on either cpu or gpu.

tomaarsen:
I agree, that would be an improvement. I'll try to find some time for this.

PR #363, opened 2023-04-19
Normalize device to CPU when evaluating

tomaarsen:
Based on discussion in https://github.com/huggingface/setfit/issues/339#issuecomment-1488142771

Hello!

## Pull Request overview
* Ensure that `y_pred` is always on the CPU when using a differentiable head.
* Test to ensure that `y_pred` is on CPU.

## Details
As discussed in #339, you can get odd errors in your custom evaluation functions when switching to using a GPU or not. This PR simplifies that a bit by always putting a torch `y_pred` on the CPU.
I believe that Transformers goes one step further, and it calls `y_pred.cpu().numpy()` (perhaps with a `.detach()` too). To me, that would also be a reasonable alternative. However, for now I've just moved it to the CPU.

- Tom Aarsen

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 import evaluate
 import numpy as np
 import torch
-from datasets import DatasetDict
+from datasets import Dataset, DatasetDict
 from sentence_transformers import InputExample, losses
 from sentence_transformers.datasets import SentenceLabelDataset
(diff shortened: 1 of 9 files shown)


src/setfit/trainer.py
```diff
@@ -3,6 +3,7 @@
 
 import evaluate
 import numpy as np
+import torch
```
alvarobartt, line 6: Hi @tomaarsen, maybe you could place the `torch` import inside the `TYPE_CHECKING` block as described my following comment

PR #390, opened 2023-06-26
Column name validation for apply column mapping

franz101:
Currently if the key is not in the dataset it fails. Here is a proposed fix for this issue.


tomaarsen:
Hello!

We have the `_validate_column_mapping` method to validate the column mapping:
https://github.com/huggingface/setfit/blob/b503c0b80b1fce800380e8e9092ab0ad14ae8889/src/setfit/trainer.py#L140-L174

This method already has several warnings to guide the user.

Perhaps you could describe the situation where `_apply_column_mapping` fails?

- Tom Aarsen

franz101:
Actually I see in this workflow it gets validated. I was using setfit for inference and applying the column mapping, so it was throwing an error.

PR #396, opened 2023-07-18
Update data.py

grofte:
If a column has been defined as a ClassLabel then `sample_dataset` strips that information away and you lose `.names`.

Test code

```python
import datasets

dataset: datasets.DatasetDict = load_dataset("SetFit/sst5")
dataset = dataset.class_encode_column("label_text")

train_dataset: datasets.Dataset = sample_dataset(dataset["train"], label_column="label_text", num_samples=8)
eval_dataset: datasets.Dataset = dataset["validation"]

model = SetFitModel.from_pretrained("sentence-transformers/paraphrase-mpnet-base-v2")

# Create trainer
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    metric="accuracy",
    batch_size=16,
    num_iterations=20, # The number of text pairs to generate for contrastive learning
    num_epochs=1, # The number of epochs to use for contrastive learning
    column_mapping={"text": "text", "label_text": "label"} # Map dataset columns to text/label expected by trainer
)

# Train and evaluate
trainer.train()
metrics = trainer.evaluate()

# Run inference
preds = model(["i loved the spiderman movie!", "pineapple on pizza is the worst 🤮"])

print(preds)
print(list(map(lambda x: train_dataset.features["label_text"].names[x], preds)))
```

I'm sure there's a better way of mapping predictions back but I am not that familiar with HuggingFace or even PyTorch.

You can run the test code without the changes and see that `train_dataset.features["label_text"]` is a Value class and doesn't have the `.names` attribute.

### DIFF
--- a/src/setfit/data.py
+++ b/src/setfit/data.py
@@ -164,10 +164,6 @@ def create_samples(df: pd.DataFrame, sample_size: int, seed: int) -> pd.DataFram
 
 def sample_dataset(dataset: Dataset, label_column: str = "label", num_samples: int = 8, seed: int = 42) -> Dataset:
     """Samples a Dataset to create an equal number of samples per class (when possible)."""
-    # Conserve ClassLabel feature if present
-    is_class_encoded = isinstance(dataset.features[label_column], ClassLabel)
-    if is_class_encoded:
-        class_label = dataset.features[label_column]
     shuffled_dataset = dataset.shuffle(seed=seed)
 
     df = shuffled_dataset.to_pandas()
@@ -177,9 +173,7 @@ def sample_dataset(dataset: Dataset, label_column: str = "label", num_samples: i
     df = df.apply(lambda x: x.sample(min(num_samples, len(x))))
     df = df.reset_index(drop=True)
 
-    all_samples = Dataset.from_pandas(df)
-    if is_class_encoded:
-        all_samples = all_samples.cast_column(label_column, class_label)
+    all_samples = Dataset.from_pandas(df, features=dataset.features)
     return all_samples.shuffle(seed=seed)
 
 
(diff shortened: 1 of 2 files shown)


grofte:
To be honest it might have been better to copy all the `.features` before converting to pandas and then reapplying them again once the transformations are done. Would look cleaner too and be more future-proof.

grofte:
For my own usage I have now wrapped `sample_data` like this:

```python
from setfit import sample_dataset

def conserved_sampling(dataset: datasets.Dataset, label_column: str, num_samples: int) -> datasets.Dataset:
    """Sample a dataset such that the feature classes are conserved."""
    features = dataset.features.copy()
    dataset_sample = sample_dataset(dataset, label_column=label_column, num_samples=num_samples)
    for feature, feature_type in features.items():
        dataset_sample = dataset_sample.cast_column(feature, feature_type)
    return dataset_sample
```

You could use this inside `sample_data` instead of what I wrote yesterday.

tomaarsen:
Hello!

I think this is a really great idea, and definitely an oversight in the original implementation. I've discovered that the [`from_pandas`](https://huggingface.co/docs/datasets/v2.14.0/en/package_reference/main_classes#datasets.Dataset.from_pandas) accepts an optional `features` parameter, so we can use that with the original `dataset.features`.

I've locally prepared this, and it's looking good. I'll push it to this PR if that's okay with you - that way you still get credited for the eventual merged commit, well deserved for spotting this oversight.

- Tom Aarsen

grofte:
Aw, I totally forgot to write a test for this. And it looks much better what you did Tom - thanks!

PR #397, opened 2023-07-19
Add missing type-hints in `SetFitTrainer`

alvarobartt:
Hi to whoever is reading this! 🤗 

## What's in this PR?

This PR adds some missing type-hints in `SetFitTrainer`, as spotted while playing around with `setfit` with @gabrielmbmb


tomaarsen:
These are some strange errors... Not sure why it does work on 3.7. Can't easily reproduce this as it uses CUDA for me locally.

tomaarsen:
I still can't reproduce it when forcing the model to CPU before sending it through openvino.

alvarobartt:
Hmm true, just changed the type-hints but didn't look at the errors, I can dig deeper if needed, just let me know if you need a helping hand 😄 @tomaarsen

tomaarsen:
Fixed the CI issues, thanks for these code quality improvements!

Issue #365, opened 2023-04-20
Questions: How can I measure overfitting? How can I know which validation examples have not been successful?


marmolpen3:
Hi, thanks for sharing SetFit. 

I am using this framework to solve a specific task in my master thesis. I would like to somehow analyze the results obtained and especially to make sure that the model is not overfitting. I think by using an epoch we reduce the chances of overfitting.

I hope to hear your response. 
Thank you very much in advance.

kgourgou:
Hello! 

I've had to think along similar lines for my work with SetFit and I think the usual data-hygiene rules apply. Are you able to split your data in parts, e.g., train/val/test, or compute metrics over cross-validations for both training and validation splits? What do the final / summary metrics look like as a function of the `num_iterations` parameter?

I also found that it helps to visualize the embeddings before and after finetunning, e.g., with TSNE, to see if the contrastive objective severely changed the structure. 

Hopefully those are somewhat useful.

tomaarsen:
You could technically, but only through
```python
num_epochs = 5
trainer = SetFitTrainer(
    model=model,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    num_epochs=1,
    ...,
)

for epoch in range(5):
    trainer.train()
    train_metrics = trainer.evaluate(train_dataset)
    eval_metrics = trainer.evaluate()
    print(f"train metrics at epoch {epoch}:", train_metrics)
    print(f"eval metrics at epoch {epoch}:", eval_metrics)
```

`trainer.model_body.encode` is exactly right, well done on spotting that. You can indeed use that before and after training to visualize embeddings. In fact, that's exactly something I've done in the past. This was a visualization of my final findings:

![image](https://github.com/huggingface/setfit/assets/37621491/93e7d5ee-c26a-4980-ae94-4fe34229df22)

In short, the training accuracy went to essentially 100%, while the evaluation performance did not. These are plotted using t-SNE.

- Tom Aarsen

marmolpen3:
How interesting, could you share the source code with the resolution or the solution of a similar example?

Thank you very much,
Elena

kgourgou:
Hi @marmolpen3 , can you be a bit more specific?

grofte:
@tomaarsen 

You can't do
> ```python
>    train_metrics = trainer.evaluate(train_dataset)
> ```

It's not an argument on the trainer class evaluate method. At least not on version 0.7.0

[...]

PR #402, opened 2023-07-26
Allow other datasets in `trainer.evaluate()`

grofte:
In issue https://github.com/huggingface/setfit/issues/365 you mention passing a different dataset to the `evaluate` than the one defined in `eval_dataset=` when you make an instance of `SetFitTrainer`. But you can't do that. Maybe you could before?

So I've added an argument and a test and otherwise only changed the code a little bit. But then I created two commits because I did it all in the browser and I don't know how to rebase (or even edit and save more than one file per commit).

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
 import evaluate
 import numpy as np
-from datasets import DatasetDict
+from datasets import Dataset, DatasetDict
 from sentence_transformers import InputExample, losses
 from sentence_transformers.datasets import SentenceLabelDataset
 from sentence_transformers.losses.BatchHardTripletLoss import BatchHardTripletLossDistanceFunction
@@ -19,7 +19,6 @@ from .utils import BestRun, default_hp_space_optuna
 
 if TYPE_CHECKING:
     import optuna
-    from datasets import Dataset
 
     from .modeling import SetFitModel
 
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -70,7 +70,9 @@ class SetFitTrainerTest(TestCase):
 
     def test_trainer_works_with_alternate_dataset_for_evaluate(self):
         dataset = Dataset.from_dict({"text": ["a", "b", "c"], "label": [0, 1, 2], "extra_column": ["d", "e", "f"]})
-        alternate_dataset = Dataset.from_dict({"text": ["x", "y", "z"], "label": [0, 1, 2], "extra_column": ["d", "e", "f"]})
+        alternate_dataset = Dataset.from_dict(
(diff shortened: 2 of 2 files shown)


tomaarsen:
Hey! Thanks for this feature!

> Maybe you could before?

You couldn't - but the Transformers project has a similar interface, and there it is possible.
I made some small changes to prevent some errors:
1. add missing `Dataset` import
2. run `make style` to run automatic formatting

Your implementation & tests look great. I'll merge this once the tests pass.

- Tom Aarsen

grofte:
Bugger, I thought I was clever enough but I still managed to forget the Dataset import. 

Next time I will download the whole thing and do all the things, I promise.

grofte:
I don't think I have much to I want to change now. 

When you add the option for multiple heads, I think a Gaussian mixture model (GMM) head trained with SGD as outlined here would be amazing: https://link.springer.com/article/10.1007/s11063-021-10599-3  
It should basically allow you to estimate whether the data you want to predict on is "in distribution" or "out of distribution". Which again tells us whether we can reasonably expect the model to give a good answer.

If you are completely unfamiliar with GMMs here's a great blog entry: https://koaning.io/posts/feed-forward-posteriors/  
Notice that the neural implementation there doesn't look great (2017 or older), but also keep in mind that expectation maximization for fitting GMMs generally fails with high-dimensionality data.

tomaarsen:
I'm not very familiar with GMM's (although I do know that blogpost author, I spoke with him earlier today haha!), but if there's a `sklearn` implementation, then that could be used instead of the default `LogisticRegression` head, as long as the new head has `fit`, `predict` and `predict_proba` methods.

I just tried using `GaussianMixture()` as the head, and results were certainly not great - but I'm probably using it incorrectly (~1.29% accuracy compared to ~87.17% with log-reg).

- Tom Aarsen

grofte:
Haha, I didn't mean as a classifier head. You might be able to do that but I am fairly certain that is completely impossible with the sklearn implementation when your body has an output with 768 dimensions.

It was more like your own idea where you compare the tSNE for the validation and training data. But by modeling the distribution of activations / body outputs for all training data you should be able to do the same thing for each individual prediction and summarize your tSNE comparison in a single number. 

So one classification head and one training data distribution head.

tomaarsen:
Ohh, I see!

grofte:
That's why I recommended Vincent's blog post. It's only about joint posteriors, not using it as a classifier. SetFit isn't doing anything specific in the contrastive training to give you outputs that look Gaussian. And I'm not sure it should - it should only just change the body enough to allow the log reg to classify successfully, right? Because we are always risking "catastrophic forgetting" when we do these kinds of fine-tunings. And then the ability to generalise outside the few-shot training data gets compromised.

tomaarsen:
> it should only just change the body enough to allow the log reg to classify successfully, right?

That's the goal, yeah - trying to make the embeddings of sentences of the same labels similar, and dissimilar if they have different labels. If successful, logistic regression should be able to solve it without any real issues.

Issue #338, opened 2023-03-28
error export onxx with body roberta


batman-do:
![image](https://user-images.githubusercontent.com/77484083/228125960-5c31b304-493b-4332-9a08-edef60ee05bd.png)

```---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[7], line 1
----> 1 export_onnx(model.model_body, model.model_head, opset=18, output_path="models/base.onnx")

File /storage/dodx/anaconda3/envs/pipeline_sentiment/lib/python3.8/site-packages/setfit/exporters/onnx.py:227, in export_onnx(model_body, model_head, opset, output_path, ignore_ir_version, use_hummingbird)
    225 if issubclass(type(model_head), models.Dense):
    226     setfit_model = OnnxSetFitModel(transformer, lambda x: model_pooler(x)["sentence_embedding"], model_head).cpu()
--> 227     export_onnx_setfit_model(setfit_model, dummy_inputs, output_path, opset)
    229     # store meta data of the tokenizer for getting the correct tokenizer during inference
    230     onnx_setfit_model = onnx.load(output_path)

File /storage/dodx/anaconda3/envs/pipeline_sentiment/lib/python3.8/site-packages/setfit/exporters/onnx.py:97, in export_onnx_setfit_model(setfit_model, inputs, output_path, opset)
     95 setfit_model.eval()
     96 with torch.no_grad():
---> 97     torch.onnx.export(
     98         setfit_model,
     99         args=args,
    100         f=output_path,
    101         opset_version=opset,
    102         input_names=["input_ids", "attention_mask", "token_type_ids"],
    103         output_names=output_names,
    104         dynamic_axes={**dynamic_axes_input, **dynamic_axes_output},
    105     )

File /storage/dodx/anaconda3/envs/pipeline_sentiment/lib/python3.8/site-packages/torch/onnx/utils.py:506, in export(model, args, f, export_params, verbose, training, input_names, output_names, operator_export_type, opset_version, do_constant_folding, dynamic_axes, keep_initializers_as_inputs, custom_opsets, export_modules_as_functions)
    188 @_beartype.beartype
    189 def export(
    190     model: Union[torch.nn.Module, torch.jit.ScriptModule, torch.jit.ScriptFunction],
   (...)
    206     export_modules_as_functions: Union[bool, Collection[Type[torch.nn.Module]]] = False,
    207 ) -> None:
    208     r"""Exports a model into ONNX format.
    209 
[...]

PR #412, opened 2023-08-25
Fixed switched token_type_ids and attention_mask

rolshoven:
I was having the same error as mentioned in #338 where I could not export my model with model_base `stsb-xlm-roberta-base`. After some debugging, I noticed that the attention_mask and token_type_ids were switched in the function `forward` (line 50) in `setfit/exporters/onnx.py`. The error then occurs because we are trying to look up both the token_type_id embedding with index 0 and the one with index 1, but there is only one embedding in the matrix. I believe that this did not happen with other model bases because they have more than two token_type embeddings. 

However, I must confess that I was not yet able to test this fix with other models that previously worked. We should definitely do this before we merge this code. To make the code safer, I also made us of kwargs when calling `self.model_body` instead of positional arguments. In my case, I was able to export the model after this small fix.

### DIFF
--- a/src/setfit/exporters/onnx.py
+++ b/src/setfit/exporters/onnx.py
@@ -1,6 +1,7 @@
 import copy
 import warnings
 from typing import Callable, Optional, Union
+from inspect import signature
 
 import numpy as np
 import onnx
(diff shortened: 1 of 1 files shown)


rolshoven:
Just noticed that the exported onnx model does only work if we switch the attention_mask and token_type_ids of the generated dcitionary after tokenization, which is probably caused by my change. I will investigate further and report back soon.

rolshoven:
I reverted my previous changes and implemented the fix directly in function `export_onnx_setfit_model` of `src/setfit/exporters/onnx.py`. The problem was that the tokenizer dictionary keys were not ordered. I implemented a generic solution that uses the signature function from module inspect to make as few assumptions on the parameters as possible while ensuring the correct order of the input values.

rolshoven:
I forgot to include the import but now the tests should work

tomaarsen:
Perhaps we can adopt the approach from #435 for ONNX, rather than sticking with the current `export_onnx`. It seems more consistent.

PR #421, opened 2023-09-18
fix mean polling

hedrergudene:
Mean pooling is not correctly calculated.

### DIFF
--- a/src/setfit/modeling.py
+++ b/src/setfit/modeling.py
@@ -90,8 +90,7 @@ copyright = {{Creative Commons Attribution 4.0 International}}
 """
 
 #Mean Pooling - Take attention mask into account for correct averaging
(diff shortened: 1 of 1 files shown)


tomaarsen:
The `SentenceTransformer` `model_body` is responsible for the mean pooling, the `normalize` call is separate from that. SetFit shouldn't have to do any mean pooling.

tomaarsen:
I believe this is just the normalization after calculating the embeddings, inspired by https://github.com/UKPLab/sentence-transformers/blob/c5f93f70eca933c78695c5bc686ceda59651ae3b/sentence_transformers/SentenceTransformer.py#L183-L184
I'll close this as I don't have confidence that the PR is correct.

- Tom Aarsen

PR #429, opened 2023-10-17
show_progress_bar as parameter on predict and predict_prob

davidsbatista:
- I want to deploy `Setfit` in production without having the progress bar shown when any of the predict methods is called, it seems there's no explicit way to do this unless changing the logging level which `sentence-transformer`
 uses to decide on the `show_progress_bar` boolean value
  
- Adding `show_progress_bar` as a parameter on `SetFitModel.predict` and `SetFitModel.predict_prob` methods

### DIFF
--- a/src/setfit/exporters/onnx.py
+++ b/src/setfit/exporters/onnx.py
@@ -153,7 +153,7 @@ def export_sklearn_head_to_onnx(model_head: LogisticRegression, opset: int) -> o
 
     # If the datatype of the model is double we need to cast the outputs
     # from the setfit model to doubles for compatibility inside of ONNX.
-    if isinstance(onnxconverter_common.data_types.DoubleTensorType, dtype):
+    if type(dtype) == onnxconverter_common.data_types.DoubleTensorType:
         sklearn_model = Pipeline([("castdouble", CastTransformer(dtype=np.double)), ("head", model_head)])
     else:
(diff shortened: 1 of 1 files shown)


davidsbatista:
The PR is now failing due to the message from flake8

> src/setfit/exporters/onnx.py:156:8: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`

which was not something my PR introduced, I can nevertheless fix it


```
    # If the datatype of the model is double we need to cast the outputs
    # from the setfit model to doubles for compatibility inside of ONNX.
    if type(dtype) == onnxconverter_common.data_types.DoubleTensorType:
        sklearn_model = Pipeline([("castdouble", CastTransformer(dtype=np.double)), ("head", model_head)])
    else:
        sklearn_model = model_head

```

tomaarsen:
I'll tackle the unrelated quality check in a separate PR. Thanks for this!

davidsbatista:
OK, if you want I can then merge/sync this with main, just let me know when that separate quality check issue is solved/merged

tomaarsen:
No worries, I'll handle it from here. I'll merge this as soon as the tests pass (which they should), despite the quality failure.

tomaarsen:
Thanks again!

davidsbatista:
happy to have contributed :)

PR #435, opened 2023-11-09
Add notebook to demonstrate how efficiently running SetFit with ONNX

MosheWasserb:
Efficiently run SetFit Models with Optimum

PR #437, opened 2023-11-10
Patch-1

RichardScottOZ:


tomaarsen:
Thank you! I've incorporated this into #435, which (re-)adds this notebook.

- Tom Aarsen

PR #441, opened 2023-11-23
fix: make sampling more reproducible

yahiaelgamal:
Without setting the `random_state` in the GroupedDataFrame.sample, we will never be able to reproduce results. Other sampling logic through the file correctly sets the random_state/seed, but this one is missing.


tomaarsen:
Hello!

Well spotted! Thanks for this :) 

- Tom Aarsen

Issue #355, opened 2023-04-13
ONNX conversion of multi-ouput classifier


jackiexue1993:
Hi,
I am trying to do onnx conversion for multilabel model using the multioutputclassifier
`model = SetFitModel.from_pretrained(model_id, multi_target_strategy="multi-output")`.
When I tried `export_onnx(model.model_body,
            model.model_head,
            opset=12,
            output_path=output_path)`, it gave me an error indicating there's no `coef_`, I understand there's no coef_ in the multioutput classifier, but is there a way to do onnx conversion for this model? 
Thanks!

tomaarsen:
Hello @jackiexue1993,

I don't have much experience with ONNX myself, but it seems like the branch from https://github.com/huggingface/setfit/pull/361 might resolve this problem. @bogedy can likely back me up on this.

You can try it out using:
```
pip install git+https://github.com/bogedy/setfit.git@onnx_change
```

- Tom Aarsen

bogedy:
Hi @jackiexue1993, yeah this ought to fix it. Let me know if you try it out and if it works.

jackiexue1993:
Hi, 
@bogedy @tomaarsen Thank you for providing this solution. I tried and it works well for the multi-output classifier. Thank you!

>

jackiexue1993:
@bogedy I double check the onnx model prediction, the accuracy was down by 5%.

bogedy:
Hmm that's troubling. Thanks, will look into.

jackiexue1993:
@bogedy I think this is a common issue for using the mpnet model. https://github.com/huggingface/setfit/issues/291#issue-1562241805

Issue #291, opened 2023-01-30
ONNX export for multilabel - discrepancy between pytorch and onnx outputs


nishanthcmesh:
Hi,
I was trying out exporting a multilabel setfit model with onnx and the exported model was giving different outputs as compared to the pytorch one. Am I doing something wrong in my inference code? Is multilabel onnx support not there at the moment?

Here is a [Colab](https://colab.research.google.com/drive/1dfR5lh1d0kJ0Sl5BuhtIclcqmJI8FEz9?usp=sharing) to recreate the issue.

nishanthcmesh:
UPDATE: I see that there is a ```torch.sigmoid``` layer added to the outputs for multitarget in ```SetFitHead```. Does this mean I will just have to add the same and check for >0.5 ?

tomaarsen:
Hello!

You're not doing anything wrong I believe, this simply does not work (yet). I ran into this a while ago when adding some more tests for the ONNX exporter, see [here](https://github.com/huggingface/setfit/blob/main/tests/exporters/test_onnx.py#L57). Perhaps @nbertagnolli, who authored the ONNX support for SetFit, can have a look at this?

- Tom Aarsen

nbertagnolli:
Let me take a look! Sorry that it wasn't working for you : (.

nbertagnolli:
I've taken a look and it seems like there are some differences between the predictions when converting mpnet based models to ONNX. I'm not sure this is an issue with the SetFit implementation of the ONNX process.  If you convert just the "sentence-transformers/paraphrase-mpnet-base-v2" model to ONNX using torch.onnx.export there are numerical differences in the output.  To see this try running with a non mpnet model like `sentence-transformers/all-MiniLM-L6-v2`.

Some code below to play with this if you'd like to try for yourself.

`from setfit import SetFitModel
from sentence_transformers import SentenceTransformer
from setfit import SetFitHead, SetFitHead, SetFitModel
from setfit.exporters.onnx import export_onnx
from setfit.exporters.utils import mean_pooling
from transformers import AutoTokenizer
import torch
import onnxruntime

model_id = "sentence-transformers/paraphrase-mpnet-base-v2"
model_id = "sentence-transformers/all-MiniLM-L6-v2"
model_body = SentenceTransformer(model_id)
model_head = SetFitHead(in_features = model_body.get_sentence_embedding_dimension(), out_features = 4, multitarget = True)

[...]

PR #446, opened 2023-11-27
Defining arguments names to avoid issues with positional args

### LINKED ISSUE #418: Inconsistent ONNX Export with Differentiable Head
Hello,

I am trying to fine-tune SetFit for a multi-class classification problem.

Everything is smooth until exporting to Onnx. The head is not exported correctly, so when loading the model with Onnx, the predictions are the outputs of some previous layer.

For example, I showcased this on a dummy dataset with 3 classes.

See https://colab.research.google.com/drive/19EESqbIDwD5FOI2Ufx22txFZ8qADq2xj?authuser=1#scrollTo=vl6AvQKqtU-9

This differs from the behaviour in the example notebook, where the LogisticRegression head is used.

Any directions would be appreciated; Otherwise, I would happily contribute with an MR if anyone can spot the issue.

pedrogengo:
Fixes #418, #291, #355

Some models that don't use `token_type_ids` were presenting different results when you run it using ONNX. This issue was due to not explicitly saying the argument names during the call, which makes the model assumes the argument was a different one.

@tomaarsen

### DIFF
--- a/tests/exporters/test_onnx.py
+++ b/tests/exporters/test_onnx.py
@@ -43,7 +43,7 @@ def test_export_onnx_sklearn_head(model_path, input_text):
             padding=True,
             truncation=True,
             return_attention_mask=True,
-            return_token_type_ids=False,
+            return_token_type_ids=True,
             return_tensors="np",
         )
         # Map inputs to int64 from int32


pedrogengo:
@tomaarsen should we include the softmax operation inside the ONNX model? Or returning the logits is the desired behavior?

tomaarsen:
I quite like the simplicity of this fix, and I do hope that this indeed resolves the problem. However, the tests seem to be problematic: 
`onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid Feed Input Name:token_type_ids`

Perhaps ONNX doesn't want `token_type_ids` to be passed, even if it is `None`? Then a solution might be:

```py
inputs = {
    "input_ids": input_ids,
    "attention_mask": attention_mask,
}
if token_type_ids is not None:
    inputs["token_type_ids"] = token_type_ids
hidden_states = self.model_body(**inputs)
```

- Tom Aarsen

pedrogengo:
Don't run the CI yet, Im still making some tests

pedrogengo:
@tomaarsen the issue with the tests is a little bit harder: for models that uses `token_type_ids` the tests run like a charm. However, for models that don't use `token_type_ids` the ONNX generates the graph without the `token_type_ids` as input. I tried your suggestion, but no success.

It is also an issue on transformers https://discuss.huggingface.co/t/deberta-v2-onnx-with-pipeline-does-not-work/35748

What I'm trying to do is to force the `token_type_ids` to appear on the graph, let's see if it works or not.

Do you have any other suggestion?

tomaarsen:
Hmm, that is tricky. I'm not extremely familiar with the inner workings of ONNX, so I don't have great suggestions I'm afraid.

- Tom Aarsen

pedrogengo:
I will try more things today, but by now, I'm showing a message when the model doesn't use the `token_type_ids`, just to validate if the whole flow is working as expected.

pedrogengo:
@tomaarsen I found a way to keep the same interface we are using today and force the `token_type_ids` to appear. It can't be an optional argument, because one time you have the ONNX graph defined, you must fill all the inputs used during the export, which means that we need to always pass the `token_type_ids` even if it is not used by the model.

On one hand it can look like: "Why keep an argument that is not used by some models?", but the answer is the generalization. Keeping the parameter for all the cases make it possible to have an export code that is general enough for all the model.

Let me know WDYT about the solution. Maybe in the feature we can work to create better interfaces for the export, but I tried to keep this PR as simple as possible.

tomaarsen:
Apologies for not responding sooner, I've been a bit busy and ONNX wasn't very high on my TODO list. Do you suspect that this PR indeed fixes the reported discrepancies? E.g. does the script from #291 behave as expected now? The fix seems so odd to me, haha.

Also, models with a differentiable head are trained a bit differently in SetFit v1.0.0, i.e. no more `freeze` and `unfreeze` calls, and only calling `trainer.train()` once. 

- Tom Aarsen

pedrogengo:
I will run the script with this branch, but on my tests I was seeing discrepancy between the results too, and after the fix it worked and returned the same scores.

Give me the weekend to see the code for v1.0.0 and I can answer here.

PR #483, opened 2024-02-01
Optimum-Intel notebook:  fix quantization explanation 

danielkorat:
Hi @tomaarsen , this PR makes changes only to a markdown cell will description of quantization.


tomaarsen:
All good! Thanks for the changes.

PR #488, opened 2024-02-08
Switch to differentiable head, larger input sequence

danielkorat:


danielkorat:
Closed due to wrong commit authoring, opened #489

PR #496, opened 2024-02-22
Fix Errors in `setfit-onnx-optimum` Notebook

danielkorat:
Hey

I found a workaround to the errors when exporting using `optimum-cli`:
I run ONNX exporting + optimization programmatically instead of using the CLI, and there are no issues.
I also replaced the SetFit model with a faster one, which uses a SetFitHead.

cc @tomaarsen @MosheWasserb @orenpereg


danielkorat:
@tomaarsen 
I added the final plot

tomaarsen:
Apologies for the delay.

Issue #578, opened 2024-12-21
ContrastiveDataset generates pair of same sentence by default. 


DemirTonchev:
What is the reason that replacement is set to True by default? Also there is no way to control that.

https://github.com/huggingface/setfit/blob/146c7c9dacdc7dca678b2fffff8ddeb79dd762c2/src/setfit/sampler.py#L15

Using this setting we get pairs such as:
("The movie was awesome", "The movie was awesome") as a positive pair in the training process. How is this beneficial?

PR #579, opened 2024-12-25
Refactor ContrastiveDataset and ContrastiveDistillationDataset 

DemirTonchev:
Refactor of ContrastiveDataset and ContrastiveDistillationDataset to generate pairs lazily. Also the trainer code is updated to create dataset from iterator. This targets how pairs are generated more motivation - #578 

```dataset = Dataset.from_generator(data_sampler.__iter__)``` this change allows to work with arbitrary big dataset (although the trade off is the cache on the disk managed by arrow dataset) 

This also targets the ContrastiveDistillationDataset bug in #578 

fixes: #578

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,6 +12,9 @@ on:
       - v*-pre
   workflow_dispatch:
 
+env:
+  TRANSFORMERS_IS_CI: 1
+
 jobs:
 
   test_sampling:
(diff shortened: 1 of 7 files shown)


tomaarsen:
Hello!

I did a bit more tests, and I'm running into an issue: If I have a dataset with no positive pairs at all, then the generator will loop indefinitely, as it keeps going until it finds e.g. 3 positive pairs.

This was also causing infinite loops via the `test_raise_when_metric_value_is_invalid` tests.

- Tom Aarsen

DemirTonchev:
Fixed. 
This occurred only when using num_iterations.

DemirTonchev:
Fixed the same problem when using sampling strategies.

PR #599, opened 2025-04-26
Ensure consistentcy between labels and probabilities

### LINKED ISSUE #587: predict and predict_proba give conflicting predictions
This is somewhat related to #462.

**The problem:**

If I call `model.predict()`, I get a label. If I call `model.predict_proba()` and try to get the final string via `model.labels` or `model.id2label`, I get a different label.

**Why this happens:**

I think it's because the sklearn model head (where the probabilities are extracted from) might reorganize the labels, leading to an inconsistency between what the model thinks the order is and how the probs are returned.

**Workaround**

Doing `model.model_head.classes_[int(np.argmax(probs))]`. However, I think that's not the intended way since classes_ is a private attribute?

emilioMaddalena:
Fixes #587.

**Problem**: In certain cases (see issue), self.labels and self.model_head.classes_ are not in the same order. When calling self.predict_proba, the returned probabilities are ordered as per the model's head labels. This however contradicts the model's attribute order and is very counter intuitive.

**Solution**: I'm simply reordering the probabilities to ensure consistency. This completely solved the problem I had described in my issue.


emilioMaddalena:
Hi @tomaarsen! Let me know if you've liked the solution or not. There are ofc multiple ways of tackling this...

PR #603, opened 2025-05-15
Address deprecated ModelCardCallback from sentence_transformers

blawney:
Addresses a minor issue related to the deprecation of `sentence_transformers.model_card.ModelCardCallback`. This has since been changed to `sentence_transformers.model_card.SentenceTransformerModelCardCallback` and the PR will catch either object type.

I initially encounted a bug when I had `setfit==1.1.1` with `sentence-transformers==4.1.0` (which resulted from an "unversioned" mamba-based build). Prior to #595 (as is the case for v1.1.1), there was an instance of `sentence_transformers.model_card.SentenceTransformerModelCardCallback` which was added by the `add_model_card_callback` method in the `sentence_transformers.SentenceTransformerTrainer` class. Since the object type did not match `sentence_transformers.model_card.ModelCardCallback` (aliased as `STModelCardCallback` in `setfit/trainer.py`), the callback was *not* removed and a call to the `setfit.Trainer` constructor resulted in this:
```
TypeError: SentenceTransformerModelCardCallback.on_init_end() missing 1 required positional argument: 'trainer'
```

Note that this is no longer an issue with `setfit==1.1.2`, yet this PR addresses the potential where either the current or deprecated callback makes its way into `setfit.trainer.BCSentenceTransformersTrainer`

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -7,7 +7,8 @@ from datasets import Dataset, DatasetDict
 from packaging.version import parse as parse_version
 from sentence_transformers import SentenceTransformerTrainer, losses
 from sentence_transformers.losses.BatchHardTripletLoss import BatchHardTripletLossDistanceFunction
-from sentence_transformers.model_card import ModelCardCallback as STModelCardCallback
+from sentence_transformers.model_card import ModelCardCallback as DeprecatedSTModelCardCallback
+from sentence_transformers.model_card import SentenceTransformerModelCardCallback
 from sentence_transformers.training_args import BatchSamplers, SentenceTransformerTrainingArguments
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thank you for the PR and the clear write-up. #647 moved all version dependent Sentence Transformers imports into a `setfit/compat.py` module, which imports `SentenceTransformerModelCardCallback` from the new `sentence_transformers.sentence_transformer.model_card` module on v5.4+, from `sentence_transformers.model_card` on v4 and v5.3, and falls back to the old `ModelCardCallback` name on v3. Combined with the `add_model_card_callback` override from #595, that covers the situation you hit. Closing this as superseded, the change will be in the upcoming v1.2.0 release.

- Tom Aarsen

PR #605, opened 2025-06-01
Update trainer.py

smartIU2:
Underlying SentenceTransformerTrainer needs fp16 flag on initialization already, to properly setup amp.

### DIFF
--- a/src/setfit/trainer.py
+++ b/src/setfit/trainer.py
@@ -49,7 +49,7 @@ class BCSentenceTransformersTrainer(SentenceTransformerTrainer):
         self.logs_prefix = "embedding"
         super().__init__(
             model=setfit_model.model_body,
-            args=SentenceTransformerTrainingArguments(output_dir=setfit_args.output_dir),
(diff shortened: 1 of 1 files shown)


smartIU2:
Hello Tom,

Thanks for taking your time to review my commit.

You're correct in that the fp16 argument gets propagated from the setfit Trainer to the sentence-transformers Trainer.

But, this argument is not what actually drives the use of amp. For this self.use_apex, self.use_cpu_amp & self.amp_dtype are set _based on the fp16 argument_ inside the [transformers Trainer](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py) init function starting from line 724, commented with "# Mixed precision setup". This init is called from the [sentence-transformers Trainer](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/trainer.py) init. Which gets called right _before_ _apply_training_arguments.

As far as I can tell these three underlying parameters are not adjusted by _apply_training_arguments afterwards.

Best regards,
Martin

Issue #608, opened 2025-07-14
AttributeError: 'Column' object has no attribute 'sort'


Murhaf:
The [latest release of datasets](https://github.com/huggingface/datasets/releases/tag/4.0.0) introduces an `IterableColumn` object (in IterableDataset); therefore, setfit won't work with datasets 4.0.0 as `samples` is no longer a list in [`set_widget_examples` in `model_card.py`](https://github.com/huggingface/setfit/blob/main/src/setfit/model_card.py#L301)

```python
    def set_widget_examples(self, dataset: Dataset) -> None:
        samples = dataset.select(random.sample(range(len(dataset)), k=min(len(dataset), 5)))["text"]
        self.widget = [{"text": sample} for sample in samples]

        samples.sort(key=len)
        if samples:
            self.predict_example = samples[0]
```

PR #609, opened 2025-07-14
Fix compatibility issues with datasets 4.0.0

Murhaf:
Fixes #608 by making sure that `samples` is a list

PR #611, opened 2025-07-25
Fix: ensure SetFit model card widget works with datasets 4.x Column objects

SaiSandeepKantareddy:
### Overview

This PR fixes #608 and related issues when running SetFit with `datasets` version 4.x.

#### Problem

With the release of `datasets` 4.0.0, columns such as "text" are now lazy `Column` objects, not native Python lists. This caused SetFit's model card widget code to fail with:

AttributeError: 'Column' object has no attribute 'sort'

when initializing the `Trainer`.

#### Solution

This patch ensures that the model card widget code converts the column to a Python list before sorting, making it compatible with both older and newer versions of `datasets`. This change maintains backward compatibility with `datasets` 2.x and 3.x.

#### Testing

- Tested with `datasets==3.6.6` and `datasets==4.0.0` on macOS and Linux.
- Verified that Trainer now works and the error is resolved.

#### Impact

- Fixes training workflow for all users on `datasets` 4.x.
- No impact on previous users or existing APIs.

---

Thanks to the community for reporting and troubleshooting (esp. [issue #608](https://github.com/huggingface/setfit/issues/608)). Please let me know if further changes or tests are needed!

### DIFF
--- a/src/setfit/model_card.py
+++ b/src/setfit/model_card.py
@@ -294,13 +294,16 @@ class SetFitModelCardData(CardData):
     def set_best_model_step(self, step: int) -> None:
         self.best_model_step = step
 
-    def set_widget_examples(self, dataset: Dataset) -> None:
+    def set_widget_examples(self, dataset: Dataset) -> None:        
         samples = dataset.select(random.sample(range(len(dataset)), k=min(len(dataset), 5)))["text"]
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I've merged #609 instead, but I do want to thank you for opening this issue and helping me try and get the compatibility issues resolves.

- Tom Aarsen

PR #634, opened 2026-04-28
Fix default_logdir import for transformers v5

elias-aouad:
Fix an import for better compatibility with transformers v5

### DIFF
--- a/src/setfit/training_args.py
+++ b/src/setfit/training_args.py
@@ -10,7 +10,7 @@ import torch
 from sentence_transformers import losses
 from transformers import IntervalStrategy
 from transformers.integrations import get_available_reporting_integrations
-from transformers.training_args import default_logdir
+from transformers.integrations.integration_utils import default_logdir
 from transformers.utils import is_torch_available
 
 from . import logging
--- a/src/setfit/training_args.py
+++ b/src/setfit/training_args.py
@@ -8,13 +8,17 @@ from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 import torch
 from sentence_transformers import losses
+
 from transformers import IntervalStrategy
 from transformers.integrations import get_available_reporting_integrations
-from transformers.integrations.integration_utils import default_logdir
 from transformers.utils import is_torch_available
 
-from . import logging
+try:
+    from transformers.integrations.integration_utils import default_logdir
+except ImportError:  # pragma: no cover
+    from transformers.training_args import default_logdir
 
+from . import logging
 
 logger = logging.get_logger(__name__)
 


peter-clark-moj:
would be great if someone could jump on this. given transformers versions <5.3.0 are considered to have a high danger security vulnerability.

bigabig:
Please merge this, it would be nice to be able to work with Transformers v5!

elias-aouad:
@tomaarsen I am pinging you in case you're still a maintainer of this repo. This is a one-liner PR, would it be possible to merge it ? It would help a lot !
Cheers 🍻
PS: Amazing work you did with this repo, it helps a lot :)

tomaarsen:
Apologies for the delay. I think it might be safer if we used a `try-except ImportError` as I imagine your new proposal won't work with older `transformers` versions.

- Tom Aarsen

elias-aouad:
> Apologies for the delay. I think it might be safer if we used a `try-except ImportError` as I imagine your new proposal won't work with older `transformers` versions.
> 
> * Tom Aarsen

Thank you for the reply, it's all done now :)

elias-aouad:
@tomaarsen sorry for the second ping, the PR is ready now with the try-except logic. It would be of great help if we could merge it 🙏 

Thanks a lot !

tomaarsen:
Hello!

Thank you for this PR, and apologies again for the slow turnaround. The try/except import you ended up with has landed on `main` as part of the wider compatibility change in #647, which also handles the other transformers v5 changes (`warmup_ratio`, `logging_dir`, the `report_to` default) together with huggingface_hub v1 and Sentence Transformers v6. I'm closing this one as superseded, the fix will be in the upcoming v1.2.0 release.

- Tom Aarsen

elias-aouad:
No problem, thanks a lot !

PS: congrats on the new releases both on setfit and sentence transformers, always great to follow your work :)

I was just wondering if setfit is intended to keep updated, or is it like a side project that could stop being maintained in the future?

All the best,
Elias

tomaarsen:
Thank you!

> I was just wondering if setfit is intended to keep updated, or is it like a side project that could stop being maintained in the future?

I would like to keep it up to date, but it is rather tricky to keep the older transformers/sentence-transformers etc. compatible. For example, the Training interface of transformers shifted a bit over time (e.g. slight renames like `evaluation_strategy`), so then it becomes easier for me to just support only the latest versions and not be fully backwards compatible (e.g. only the new names like `eval_strategy` would work).

Either way, I would like to keep setfit up to date, I think it's a valuable little library that I would like to keep working. I'll try to be a bit better to stay on top of the updates, as it usually doesn't take too much time to make sure it's working. Sometimes I just get distracted by Sentence Transformers, as that's my main focus overall.

- Tom Aarsen

PR #635, opened 2026-04-29
uv handles dependencies

elias-aouad:


tomaarsen:
Hello!

I don't see the immediate benefit in moving from setup.py to pyproject.toml, although I do recognize that it's better eventually/overall. But if it also comes with a lot of backwards incompatible deletions, then I just can't accept that.

- Tom Aarsen

PR #643, opened 2026-08-17
fix: honor report_to when constructing the internal ST trainer

tonycoder-hub:
## Summary
`BCSentenceTransformersTrainer` built `SentenceTransformerTrainingArguments` with only `output_dir`, so `report_to` defaulted to `"all"`. `transformers.Trainer.__init__` registered every installed integration (including DagsHub) before `_apply_training_arguments` copied the user's `report_to`.

On Windows, `DagsHubCallback.setup` then splits `MLFLOW_TRACKING_URI` on `os.sep` and crashes even when the user asked for `report_to="mlflow"` (or `"none"`).

Pass `report_to` at construction time so callbacks match the user's setting.

Closes #621

## Test plan
- [x] `pytest tests/test_trainer.py tests/test_training_args.py tests/test_model_card.py -q` → 50 passed, 2 skipped
- [x] New `test_trainer_report_to` fails on current main (`report_to="none"` still registers a dummy integration) and passes after the change

PR #644, opened 2026-08-17
fix: apply GROUP_BY_LABEL sampler for in-batch embedding losses

tonycoder-hub:
train_embeddings compared an instantiated loss module against loss classes, so loss in (BatchAllTripletLoss, ..., SupConLoss) was always False and GROUP_BY_LABEL never applied. Compare args.loss (the class) and reset the sampler between train() calls. Tests: python -m pytest tests/test_trainer.py -k batch_sampler or non_default_loss -v --no-cov -> 9 passed. Distinct from #643/#620/#579/#627.


tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #645, opened 2026-08-17
fix: use actual median for model-card word count

tonycoder-hub:
SetFitModelCardData.set_train_set_metrics labeled a row Min | Median | Max but computed the middle cell as the mean (sum/len). Every generated model card reported an average under Median.

Now uses statistics.median. Test uses counts 2, 3, 4, 11 so median 3.5 and mean 5.0 diverge.

Tests: python3 -m pytest tests/test_model_card.py -q — new test fails on main (Median 5.0) and passes after. Distinct from #643/#644.


tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #646, opened 2026-08-27
Exclude identity pairs from contrastive pair generation

### LINKED ISSUE #620: Identity pairs are included in the training
Currently the training always includes identity pairs as `shuffle_combinations(iterable: Iterable, replacement: bool = True)` has the `replacement` parameter set by default to true. This is inconsistent with the setfit paper:

> Assuming that a small number (K) of labeled examples are given for a binary classification task, the
potential size of the ST fine-tuning set T is derived
from the number of unique sentence pairs that can
be generated, namely K(K − 1)/2, which is significantly larger than just K.

Putting aside the alignment with the paper, I think it does not make sense to include these pairs in the dataset as they are already "perfectly fitted". 
Is this behaviour intentional? If yes, are there any plans for a possibility for switching the identity pairs inclusion off?

tonycoder-hub:
## Problem

`shuffle_combinations` defaults to `replacement=True`, so contrastive training always includes identity (self) pairs. Those pairs carry no training signal and disagree with the SetFit paper, which counts unique sentence pairs as \(K(K-1)/2\). See #620.

## Fix

- Call `shuffle_combinations(..., replacement=False)` for both single-label and multilabel pair generation.
- When a dataset has one sentence per label (no positives) or a single label (no negatives), clamp those pair counts to 0 so we do not sample from an empty list.

`ContrastiveDistillationDataset` is left unchanged: pairing a sentence with itself is a valid cosine-similarity = 1 target there.

Related: #579 and #627 rewrite pair generation more broadly (lazy / streaming). This is the minimal change for #620 against current `main`.

## Test plan

- Updated expected pair counts in `test_sentence_pairs_generation` and `test_sentence_pairs_generation_multilabel` (no more self-pairs).
- Added `test_sentence_pairs_generation_excludes_identity_pairs` across all sampling strategies, single- and multilabel.
- Added edge-case tests for one sentence per label and a single label.

Fixes #620


tonycoder-hub:
Friendly review request for @tomaarsen when you have a moment — excludes identity pairs from contrastive sampling (`replacement=False`) per #620, with tests. Commit authorship was cleaned up to sole author. Happy to adjust if needed. Thanks!

tonycoder-hub:
Quick follow-up: the PR checks (Unit tests / Quality / Build PR Documentation) are sitting on GitHub’s first-time contributor **Approve and run workflows** gate (`action_required`), so the rollup stays UNSTABLE until a maintainer approves them.

If someone with write access can approve the pending runs on this PR, I can rebase onto current `main` afterward if needed. Thanks!

PR #653, opened 2026-09-11
Task-aware encoding and image inputs for multimodal Sentence Transformer bodies

oneryalcin:
## Summary

SetFit can now be trained on images, not just text, when its body is a multimodal Sentence Transformers model. It also gains a `task` option for bodies whose processor insists on knowing whether an input is a query or a document.

```python
from PIL import Image
from setfit import SetFitModel, Trainer

model = SetFitModel.from_pretrained("Hcompany/NeoMME-260M-Retriever-ST-dense", task="document")
trainer = Trainer(model=model, train_dataset=ds)   # ds["text"] holds PIL.Image rows, ds["label"] the classes
trainer.train()
model.predict([Image.open("page.png")])
model.save_pretrained("my-page-classifier")        # task is saved with the model
```

Nothing changes for existing text users: `task` defaults to `None` and every code path behaves as before.

## Why

Sentence Transformers v6 embeds images through the same `encode` API as text. That means SetFit's contrastive pair sampling, batching and scikit-learn head already work on image rows in principle. In practice three things broke, and they are the three changes in this PR.

## What changed

**1. The model card assumed every training example is a string.**
`Trainer.__init__` crashed with `TypeError: object of type 'PngImageFile' has no len()` while building widget examples, and the training-set summary called `.split(" ")` on every row. Whether a dataset is text is now decided from its features (`Value("string")`). Non-text datasets get no Hub widget and no word-count table, and the usage snippet on the card shows an image call instead of a text one. The label-count table is kept.

**2. Training data could not carry a task.**
Some encoders refuse inputs without a task; NeoMME raises `chat templates require task='query' or task='document'`. Sentence Transformers resolves the task per column through `router_mapping`, which SetFit never set. The trainer now maps the pair columns `sentence_1`, `sentence_2` and `sentence` to `SetFitModel.task`, on both the training arguments and the data collator.

**3. Prediction called `encode()` without a task.**
`SetFitModel.encode`, the scikit-learn `fit` path and the distillation teacher now pass `task` through when it is set, via a small `encode_kwargs` property. When it is not set, nothing is passed, so bodies on older Sentence Transformers versions never see an unknown keyword.

## Behaviour details

- `task` is a new argument on `SetFitModel` and `from_pretrained`, saved in `config_setfit.json`, also for `SpanSetFitModel`, so a reloaded model keeps it.
- The routing is written into the Sentence Transformers arguments before the trainer builds its data collator, so bodies with a `Router` module accept a task instead of failing at initialization.
- When `model_init` swaps in a model without a task, the routing is reset to the empty default.
- Requires sentence-transformers v5.0.0, where `task` and `router_mapping` were introduced; older versions get a clear `ValueError`.
- Rejected with `NotImplementedError` for the differentiable head, whose dataset tokenizes inputs as text.

## Tests

- `test_trainer_image_inputs`: trains, predicts, checks model-card fields and save/reload on `PIL.Image` rows with `hf-internal-testing/tiny-random-CLIPModel` (a few MB). Skipped below sentence-transformers v6.
- `test_trainer_routes_pair_columns_through_task`: router mapping is set only when a task is given and cleared when a task-less model is swapped in.
- `test_trainer_with_router_body`: a `Router` body trains with `task="document"` on `sentence-transformers-testing/stsb-bert-tiny-safetensors`.
- `test_task_forwarded_to_encode`, `test_task_persists_across_save_load`, `test_task_persists_for_span_model`, `test_task_rejects_differentiable_head`, `test_task_requires_sentence_transformers_v5`.
- `test_save_load_config` updated for the new `"task": null` key.

Full suite locally: 153 passed, 15 skipped. The 3 failures (`test_model_card`, `test_differentiable_models[True-*]`) fail identically on `main` in the same environment (torch 2.14 / transformers 5.17). Run on sentence-transformers 6.0.1 / transformers 5.17.0, and on sentence-transformers 5.7.0 / transformers 4.57.6 where the image test skips as intended.

Also verified outside the suite on `Hcompany/NeoMME-260M-Retriever-ST-dense` with rendered PDF pages: `from_pretrained(..., task="document")` and no other changes trains, predicts, saves and reloads.

## Docs

New how-to page `docs/source/en/how_to/image_inputs.mdx`, and a `task` entry in the `from_pretrained` docstring.

## Known, pre-existing

Swapping the model through `BCSentenceTransformersTrainer.setfit_model` leaves the data collator's `preprocess_fn` bound to the previous body. This predates the change and is unchanged here.

## Out of scope

Differentiable head with images, ABSA and distillation on image inputs, and late-interaction (multi-vector) bodies.

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -27,6 +27,8 @@
     title: Classification Heads
   - local: how_to/multilabel
     title: Multilabel Text Classification
+  - local: how_to/image_inputs
+    title: Image Classification
   - local: how_to/zero_shot
     title: Zero-shot Text Classification
   - local: how_to/hyperparameter_optimization
--- a/docs/source/en/how_to/image_inputs.mdx
+++ b/docs/source/en/how_to/image_inputs.mdx
@@ -1,0 +1,59 @@
+
+# Image Classification
+
+SetFit can train few-shot classifiers on images, not only text, by using a multimodal Sentence Transformer body.
+Sentence Transformers v6 embeds images through the same `encode` API as text, and SetFit's contrastive training and
+classification head sit on top of that API unchanged.
+
+<Tip>
+
+This requires `sentence-transformers[image]>=6.0.0` and a body whose processor accepts images, for example a CLIP
+checkpoint or a multimodal document encoder such as `Hcompany/NeoMME-260M-Retriever-ST-dense`. `model.modalities`
+reports what a loaded body accepts.
+
+</Tip>
+
+## Passing images
+
+The `"text"` column of the training dataset holds the inputs, whatever their modality. `PIL.Image` objects, local
+paths and URLs are all accepted by Sentence Transformers; a 🤗 Datasets `Image` feature decodes to `PIL.Image` rows
+automatically.
+
+```py
+from datasets import Dataset
+from PIL import Image
+
+from setfit import SetFitModel, Trainer, TrainingArguments
+
+train_dataset = Dataset.from_dict({
+    "text": [Image.open("invoice_1.png"), Image.open("contract_1.png"), ...],
+    "label": ["invoice", "contract", ...],
+})
+
+model = SetFitModel.from_pretrained("sentence-transformers/clip-ViT-B-32")
+trainer = Trainer(model=model, args=TrainingArguments(num_iterations=10), train_dataset=train_dataset)
+trainer.train()
+
+model.predict([Image.open("invoice_7.png")])
+# ['invoice']
+```
+
+## Bodies that require a task
+
+Some multimodal encoders, such as document retrievers, require every input to be embedded under a task like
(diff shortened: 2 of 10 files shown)


oneryalcin:
Pushed d7f152f after a scaling failure on a real run: fine-tuning on 3,887 page images built 7.5M candidate pairs and embedded ~78K PIL images into one Arrow table, which failed with `offset overflow while concatenating arrays` after an hour on an H100.

Two changes, text behaviour unchanged:

1. `ContrastiveDataset` stops enumerating once `num_iterations * n` positive and negative pairs exist (only that prefix was ever served). Verified identical pairs vs. upstream for text inputs.
2. For non-str inputs, pair rows hold indices into `Trainer._pair_inputs`; the ST collator's `preprocess_fn` (`tokenize_fn` on ST 5) resolves them at batch time. Pair build for 4,972 images: 0.4 s, dataset is int64 columns.

Tests: `test_num_iterations_bounds_pair_generation`, `test_get_dataset_keeps_images_out_of_pair_table`, plus an assertion in `test_trainer_image_inputs`.

